PR #23, opened 2018-06-04
call dot command with temporary file , graph name in dot file must be quoted

kenichiro-kimura:
1. for FreeBSD-amd64, the dot command must be called with temporary file, not with pipe and stdin
2. in dot file, the graph name must be quoted( to use '-' in graph name)


cmd/templates.go
```diff
@@ -0,0 +1,29 @@
+package dot
```
k1LoW: Could you remove this file ? ( may be unnecessary )

kenichiro-kimura: Oh,sorry.. I removed the file. Thanks!

k1LoW:
Thank you!!! :+1:

PR #136, opened 2019-08-19
add Dockerfile

peccu:
Thank you for making cool tool for DB.
I think this tool is useful with docker, so I wrote simple Dockerfile.

with README.md updated, but it should work after docker hub configuration.
( Why the tag `k1low/tbls` is lowercase that is because from docker's restriction `invalid reference format: repository name must be lowercase`.)

You can use in `docker-compose.yml` too like this.

```yml
version: '3.7'
services:
  tbls:
    # from docker hub
    # image: k1low/tbls

    # from current docker/Dockerfile
    build: ./docker

    depends_on:
      - db_service
    volumes:
      - .:/work
    command: "doc postgres://dbuser:dbpass@db_service:5432/dbname"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,12 @@ Document a database with one command.
 $ tbls doc postgres://dbuser:dbpass@hostname:5432/dbname
 ```
 
+Using docker image.
+
+```console
+$ docker run --rm -v $PWD:/work k1low/tbls doc postgres://dbuser:dbpass@hostname:5432/dbname
+```
+
 ## Install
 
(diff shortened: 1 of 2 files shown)


k1LoW:
@peccu !!! Great! Thank you for your commit !!!

> but it should work after docker hub configuration

OK!!

k1LoW:
Create https://cloud.docker.com/repository/docker/k1low/tbls !

peccu:
Thank you for merging and configuring docker hub 👏

Issue #173, opened 2020-03-02
show public objects explicitly?


ykostsik:
Hi @k1LoW,

It's a followup of #158 
the `tbls lint` returns the full object name (schema_name.table_name[.column_name]) for the objects except the public schema. It just omits the `public` word in the output. 
That's why it's impossible to filter the public objects out just using 
```
Exclude:
 - public.*
```
I just wonder, is it difficult to add this one into code?
thanks in advance.

k1LoW:
fmm.

>  It just omits the public word in the output.

Indeed tbls omits the word public.

https://github.com/k1LoW/tbls/blob/fc2f87e1ecf6f2b15cf2c52e6bd2820facee272f/drivers/postgres/postgres.go#L64-L66

Give me time to think..

## Workaround

tbls (master branch) supports [`include:` option](https://github.com/k1LoW/tbls#filter-tables). If it works in your case, I will release it.

ykostsik:
I need to test this first. But yes, it seems like a one more way to simplify the config file when dealing with a huge number of the DB objects.

k1LoW:
Hi @ykostsik . tbls show `public.` at v1.28.0.

ykostsik:
Hi @k1LoW, thanks for a quick fix implemented. I suppose it will be correct to show it, because of not everyone uses default schemas only. I will test it on Monday, this issue will is closed.

PR #175, opened 2020-03-06
[BREAKING][PostgreSQL][Amazon Redshift]Show `public.` schema

k1LoW:
Because `public.` is **NOT** default schema name.
( see ["5.9.3. The Schema Search Path"](https://www.postgresql.org/docs/current/ddl-schemas.html) )


k1LoW:
link: https://github.com/k1LoW/tbls/issues/173

PR #196, opened 2020-04-13
Fixed an error in Graphviz when the parent table of Relation is an exclude target.

yoskhdia:
In Config.FilterTables, the match with the child table of the relation was checked, but not with the parent table.
Because of this, the table specified in Exclude (or Excluded by Include) remains in the Schema.Relations property when `tbls doc` is executed.
It works until the dot file is generated, but an error like `node excluded_table, port id unrecognized` occurs when an image is generated with Graphviz afterwards.


k1LoW:
Hi @yoskhdia ! Thank you for your GREAT FIX !

Issue #213, opened 2020-04-24
Redshift: separate quoting around schema and table name breaks reference parser


mjpieters:
### What happened

I'm testing tbls against a schema with a reserved name (`time`) as well as with a hyphenated table name:

```sql
CREATE SCHEMA time;

CREATE TABLE time.bar (
  id int IDENTITY(1, 1) PRIMARY KEY
);
CREATE TABLE time."hyphenated-table" (
  id int IDENTITY(1, 1) PRIMARY KEY
);
CREATE TABLE time.referencing (
  id int IDENTITY(1, 1) PRIMARY KEY,
  bar_id int NOT NULL,
  ht_id int NOT NULL,
  CONSTRAINT referencing_bar_id FOREIGN KEY(bar_id) REFERENCES time.bar(id),
  CONSTRAINT referencing_ht_id FOREIGN KEY(ht_id) REFERENCES time."hyphenated-table"(id)
);
```

While foreign keys are not enforced by Redshift, the query optimiser does use them so it makes sense to put these in.

The above schema breaks tbls' relations parsing, because it doesn't account for the quoting that AWS redshift puts on both the schema and on the second table, when querying constraints:

```sql
select pg_get_constraintdef(oid) as def from pg_constraint
where conrelid = 'time.referencing'::regclass and contype = 'f';
                def
-----------------------------------------------------
 FOREIGN KEY (bar_id) REFERENCES "time".bar(id)
 FOREIGN KEY (ht_id) REFERENCES "time"."hyphenated-table"(id)
(2 rows)
```

Note the `"time"` quoting in the response, and the separate quotes around `hyphenated-table`.

This causes tbls to be unhappy, because the postgres driver only [strips quotes from the start and end of the string](https://github.com/k1LoW/tbls/blob/10559b552d33f6affea3a8531e29ef98ed11b74c/drivers/postgres/postgres.go#L340)

With `DEBUG=1` we thus see:

```
not found table 'time".bar'
github.com/k1LoW/tbls/schema.(*Schema).FindTableByName
	/Users/k1low/src/github.com/k1LoW/tbls/schema/schema.go:136
github.com/k1LoW/tbls/drivers/postgres.(*Postgres).Analyze
	/Users/k1low/src/github.com/k1LoW/tbls/drivers/postgres/postgres.go:359
github.com/k1LoW/tbls/datasource.Analyze
	/Users/k1low/src/github.com/k1LoW/tbls/datasource/datasource.go:93
[...]

Issue #212, opened 2020-04-23
Postgres: Support Materialized Views


mjpieters:
In Postgres, a [Materialized View](https://www.postgresql.org/docs/10/rules-materializedviews.html) is basically a cached copy of a query, and has a lot in common with views and tables. They are treated just like tables when querying, and just like views in that they have an underlying query that produced the data.

However, they are not showing up in the `tbls` documentation, because they are neither a view, nor a table, and the [current information schema query](https://github.com/k1LoW/tbls/blob/10559b552d33f6affea3a8531e29ef98ed11b74c/drivers/postgres/postgres.go#L71-L79) doesn't include materialized views anyway (they are listed in the `pg_catalog.pg_class` table but not in `information_schema.tables`).

Current materialized views can be listed with the [`pg_catalog.pg_matviews` view](https://www.postgresql.org/docs/current/view-pg-matviews.html). The following version of the `tableRows` query adds in materialized views:

```sql
SELECT DISTINCT cls.oid AS oid, cls.relname AS table_name, tbl.table_type AS table_type, tbl.table_schema AS table_schema
FROM pg_catalog.pg_class cls
INNER JOIN pg_namespace ns ON cls.relnamespace = ns.oid
INNER JOIN (SELECT table_name, table_type, table_schema
FROM information_schema.tables
WHERE table_schema != 'pg_catalog' AND table_schema != 'information_schema'
AND table_catalog = $1
UNION SELECT matviewname as table_name, 'MATERIALIZED VIEW' as table_type, schemaname as table_schema
FROM pg_catalog.pg_matviews
WHERE schemaname != 'pg_catalog' AND schemaname != 'information_schema'
) tbl ON cls.relname = tbl.table_name AND ns.nspname = tbl.table_schema
ORDER BY oid;
```

If we then update the view definition handling to use the [`pg_catalog.pg_get_viewdef()` function](https://www.postgresql.org/docs/current/functions-info.html) (which works for both views and materialized views) the code can also handle MATERIALIZED VIEWs. Note: `pg_get_viewdef()` takes an OID, but currently the [OID is stored as a string](https://github.com/k1LoW/tbls/blob/10559b552d33f6affea3a8531e29ef98ed11b74c/drivers/postgres/postgres.go#L90), this should really be a `uint64`!

This is a rough sketch of how this could work:

```go
		// (materialized) view definition
		if tableType == "VIEW" || tableType == "MATERIALIZED VIEW" {
[...]

PR #214, opened 2020-04-24
Postgres: support materialized views

mjpieters:
This is my proposed implementation for #212

I first refactored the way the postgres driver queries for table information. The original code relied mostly on `information_schema` (the SQL-92 standard system catalog) but added in joins against the Postgres-specific `pg_catalog` tables, my version uses *just* the `pg_catalog` system catalog. There isn't really much point in using `information_schema` unless you are trying to write standards-compliant SQL that is easily portable to other databases. Materialized view information is largely absent from the `infromation_schema` catalog and so not refactoring would have lead to quite convoluted joins or union queries.

This is _mostly_ resulting in the exact same output, except for array types and domains. The `information_schema` view on attribute types is quite limited and make it quite hard to get proper type info out, but this is not the case when using the `pg_attribute` and `pg_type` tables in conjunction with the `format_type()` function.

For arrays, rather than `array` you now get the contained type and dimensions, so `text[]` (array of strings), `double precision[][]` (array of arrays of floating point values), etc. For domains (which are 'reusable constraints', attaching the constraint to a type rather than to a column or table), instead of the constrained type, you get the domain name. IIRC, for certain variable-length types you'll now get the table-specific length as well where that wasn't reflected before.

I've retained the preference for `varchar(...)` over `character varying(...)`, however, applying that change at the query level.

Further changes in this refactor:

- use `tableOid` everywhere to fetch information about the table, rather than the schema and  table name strings.
- fetch comments inline when fetching tables and columns.
- use pq.Array() to fetch column names in constraints and indexes, avoiding potential parsing issues with column names containing the delimiter.

I've run the full test suite with this, as well as manually tested this agains AWS Redshift (see separate issue I found that's independent of this pull request, #213).


k1LoW:
@mjpieters THANK YOU  FOR YOUR GREAT COMMIT !!! 🚀 🚀 🚀 
I'll check it !!!

drivers/postgres/postgres.go
```diff
@@ -228,42 +210,28 @@ ORDER BY pt.tgrelid
...
-WHERE table_name = $1
-AND table_schema = $2
-ORDER BY ordinal_position
-`, tableName, tableSchema)
+SELECT
+    attr.attname AS column_name,
+    pg_get_expr(def.adbin, def.adrelid) AS column_default,
+    NOT (attr.attnotnull OR tp.typtype = 'd' AND tp.typnotnull) AS is_nullable,
+    CASE
+        WHEN attr.atttypid::regtype = ANY(ARRAY['character varying'::regtype, 'character varying[]'::regtype]) THEN
```
k1LoW, line 220: fmfm.

mjpieters: Yeah, I’m not that happy with this part. Looking over pg_type again this should really be a test for

```sql
    attr.atttypid::regtype = 'character varying'::regtype
OR (tp.typarray != 0 AND tp.typarray::regtype = 'character varying'::regtype)
```

Or, just drop the whole notion of transforming `character varying` to `varchar` altogether ;-)

Issue #232, opened 2020-05-08
Support MSSQL description


paulKabira:
Previous Title:
> [Question] Which property is read for MSSQL server comments?

First of all, great tool.

I am using this tool to document my MSSQL server and would like to know which property is used to generate the comment data. I have tried adding the widely popular `MS_Description`(set from SMSS description) property to tables and columns but the generated output contains empty cells for **Comments** column and **Description** header under table.

I see no options in configuration for specifying the property. Am I doing something wrong?
Any help is appreciated.

paulKabira:
Went through the source code, the `mssql` driver does not select table/column comments. This is officially a feature request. Now, I would have made a pull request but I don't have the slightest idea of go development workflow or the language (I _can_ make a pull request with the changes I made which worked out for me). So I will just list out the queries I used.

#### Table with comments
```sql
SELECT schema_name(schema_id) AS table_schema, k.name, k.object_id, k.type, CAST(e.value AS VARCHAR) AS table_comment
FROM sys.objects AS k
LEFT JOIN sys.extended_properties AS e ON
e.major_id = k.object_id AND e.name = 'MS_Description' AND e.minor_id = 0
WHERE type IN ('U', 'V')  ORDER BY OBJECT_ID
```

#### Columns with comments
```sql
SELECT
  c.name,
  t.name AS type,
  c.max_length,
  c.is_nullable,
  c.is_identity,
  object_definition(c.default_object_id),
  CAST(e.value AS VARCHAR) AS comment
FROM sys.columns AS c
LEFT JOIN sys.types AS t ON c.system_type_id = t.system_type_id
LEFT JOIN sys.extended_properties e ON
e.major_id = c.object_id AND e.name = 'MS_Description' AND e.minor_id = c.column_id
WHERE c.object_id = $1
and t.name != 'sysname'
ORDER BY c.column_id
```
**Note 1**: These are the same queries with an additional join for the comments.
**Note 2**: `MS_Description` is the standard extended property for comments.
Hope it helps.

*PS: Let me know if you want me to make a pull request.*

k1LoW:
@paulKabira Thank you for your report !!!

paulKabira:
The `MS_Description` gets only 30 characters from the table. This is because the value is cast to a `VARCHAR` instead of `NVARCHAR(MAX)`.
So the new SQLs are

**Table**
```sql
[...]

PR #234, opened 2020-05-08
Support MSSQL Description for Table/Columns.

paulKabira:
Add support for MSSQL Description `MS_Description` which is the
standary extended property for comments.

Fix: #232


k1LoW:
!!!!!! @paulKabira . Thank you for your GREAT commit !!!!

PR #260, opened 2020-11-03
Add flag that it should deltect relations 

syarig:
I implemented flag that it should deltect relations  

if you approve this change, I'm going to add testcases.

Could you review it.

Version
- tbls 1.43.1
- MySQL 5.7
- macOS 10.15.6

## Changes

added ShouldDetectRelations flag. If It is on, detect relations and merged.
The following is examples of automatic detection.

### Ex. 1

- table: posts
- columns
  - user_id
- parentTable: users
- parentColumns:
  - id

### Ex. 2

- table: likes
- columns
  - user_post_id
- parentTable: user_posts
- parentColumns:
  - id


k1LoW:
Hi @syarig !! I think a good feature, too.

I know that the rules for identifying a relation from `users.id` and `user_id` are the same as those of Ruby on Rails and some WAFs, but they are not common for all RDBMS users.

Is it possible to take one or the other?

1. allow for relation-detection rules
2. make the name of the flag a bit more restrictive

syarig:
Hello! @k1LoW
I'm grad to hear that it's a good feature!

Thank you for review and I improved by receiving your review.
I added concept the naming strategy. It's extensible, so could handle various cases by adding function
What do you think?

k1LoW:
> I added concept the naming strategy. 

GREAT!! Thank you!!!

I think this concept is a useful concept that can be used outside of a relationship setting.
I also think it's a setting you don't usually have to worry about.

Because most users will only care about whether or not the relationship is auto-detected and generated.

So I want to hide the NamingStrategy setting.

**Your PoC**

`strategy:` is `default`, but users should write `naming:`.

```yaml
naming:
    # strategy: default
    detectRelations: true
```

**My suggestion 1:**

Most users write only `detectVirtualRelationsByName: true`. Only users who want to do more detailed configuration can consciously write `namingStrategy:`

```yaml
detectVirtualRelationsByName: true
[...]
# namingStorategy: default
```

**My suggestion 2:**

OR

```yaml
detectVirtualRelations: 
    enable: true
    # storategy: default
```

I like `suggestion 2`.

config/naming.go
```diff
@@ -0,0 +1,70 @@
...
+type NamingStrategy struct {
+	ParentTable  Namer
+	ParentColumn Namer
+}
+
+// SelectNamingStrategy sets the naming strategy
+func SelectNamingStrategy(name string) bool {
+	switch name {
+	case "":
+	// TODO: Add case if added naming strategy
```
k1LoW, line 30: 👍

syarig:
@k1LoW Sure. You have a point. Thank you for your suggestion.  I applied Suggestion 2P and Added unit tests

k1LoW (approved):
@syarig GREAT!!!! THANK YOU!!!!! 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀

PR #261, opened 2020-11-03
Add baseUrl flag for links

wubin1989:
Hello!

I want to publish our .md files to code hosting service of aliyun, but links are not correct, because lacking of base url.
So I add a flag baseUrl to support this feature.

Usage:
`tbls doc -b xxx/schema/`

the `xxx/schema/` will be added to prefix of all links in .md files. For example: `er`.
So the er picture will show.

### DIFF
--- a/cmd/doc.go
+++ b/cmd/doc.go
@@ -198,7 +198,7 @@ func init() {
 	docCmd.Flags().BoolVarP(&withoutER, "without-er", "", false, "no generate ER diagrams")
 	docCmd.Flags().BoolVarP(&adjust, "adjust-table", "j", false, "adjust column width of table")
 	docCmd.Flags().StringVarP(&when, "when", "", "", "command execute condition")
-	docCmd.Flags().StringVarP(&baseUrl, "baseUrl", "b", "", "base url for links")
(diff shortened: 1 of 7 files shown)


cmd/doc.go
```diff
@@ -197,6 +198,7 @@ func init() {
 	docCmd.Flags().BoolVarP(&withoutER, "without-er", "", false, "no generate ER diagrams")
 	docCmd.Flags().BoolVarP(&adjust, "adjust-table", "j", false, "adjust column width of table")
 	docCmd.Flags().StringVarP(&when, "when", "", "", "command execute condition")
+	docCmd.Flags().StringVarP(&baseUrl, "baseUrl", "b", "", "base url for links")
```
k1LoW, line 201:
```suggestion
	docCmd.Flags().StringVarP(&baseUrl, "base-url", "b", "", "base url for links")
```

k1LoW (commented):
Hi @wubin1989 !! Thank you for your commitment!

Only 1 suggestion has been added.
Please check it.

cmd/doc.go
k1LoW: Or maybe the `base-url` option itself isn't necessary.

k1LoW:
@wubin1989 Thank you!!!

PR #287, opened 2021-02-01
Add section about updating the documentation

tomi:
Add documentation how to update the database documentation using the `--force` flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,14 @@ $ tbls diff postgres://dbuser:*****@local:5432/dbname postgres://dbuser:*****@pr
 
 > **Notice:** `tbls diff` shows the difference Markdown documents only.
 
+### Update database documentation
+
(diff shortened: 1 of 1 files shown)


k1LoW:
Thank you for your commit!

I think 

- `--force` flag is not for updates, it's just an option to force overwrite of the document. For example, the md file of a table that has been removed from the data source will not be removed by `--force`.
- `--rm-dist` flag removes all files in the document path. It is not a flag for updating.

This means that tbls does not currently have the features to update the document correctly yet.

k1LoW (changes requested):
https://github.com/k1LoW/tbls/pull/287#issuecomment-771268627

tomi:
I updated the README to describe these flags

k1LoW:
@tomi GREAT!!! THANK YOU!!

k1LoW (approved):

PR #293, opened 2021-02-04
Handling empty env vars in strict mode

navaneeth-spotnana:
We use `set -o nounset` and this is causing the quick install script to fail:

```
/tmp/use-tbls.tmp: line 21: TBLS_TMPDIR: unbound variable
```

Solution: https://stackoverflow.com/a/11264114/3073656


k1LoW:
@navaneeth-spotnana Nice!! Thank you for your commitment!!!

Issue #300, opened 2021-02-25
Add additional information for postgress


muhfaris:
I thought adding more information on connection PostgreSQL is better.
when using SSL mode the config like below :
```
# DSN (Database Source Name) to connect database
dsn: postgres://dbuser:dbpass@localhost:5432/dbname?sslmode=disable
```
or
```
# DSN (Database Source Name) to connect database
dsn: postgres://dbuser:dbpass@localhost:5432/dbname?sslmode=verify-full&sslrootcert=ca-cert.pem
```

some available sslmode of postgresql : disable, allow, prefer, require, verify-ca, verify-full

YauhenPylAurea:
it feels like this issue should be closed ...

PR #309, opened 2021-05-13
Add PostgreSQL SSL mode in README.md

kakisoft:
In README, there is no statement about PostgreSQL in SSL mode.
So, I would like to suggest to add it.

Refer to:
https://github.com/k1LoW/tbls/issues/300

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,12 +61,6 @@ Using docker image.
 $ docker run --rm -v $PWD:/work k1low/tbls doc postgres://dbuser:dbpass@hostname:5432/dbname
 ```
 
-In SSL mode, add "?sslmode=disable"
-For example:
-```console
-$ docker run --rm -v $PWD:/work k1low/tbls doc postgres://dbuser:dbpass@hostname:5432/dbname?sslmode=disable
-```
-
 ## Install
 
 **deb:**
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -61,6 +61,12 @@ Using docker image.
 $ docker run --rm -v $PWD:/work k1low/tbls doc postgres://dbuser:dbpass@hostname:5432/dbname
 ```
 
+In SSL mode, add "?sslmode=disable"
```
k1LoW, line 64: I think you need to disable ssl mode when the PostgreSQL server does not support SSL connections.

k1LoW:
@kakisoft Thank you!!

README.md
kakisoft: Sorry, I didn't explain enough.
I just would like you to add this statement, not replace.

So, This is what I meant.

Before:
image

After:
image

But, I made a mistake first pull request.
I'm sorry for confusing you.

I tried to add the statement under PostgreSQL connection.
But, first pull request has an error, I added the statement under Quick Start.

So, I fixed it and sent you pull request again.

k1LoW: Why `In SSL mode, add "?sslmode=disable"` ?

SSL mode is a connection method between client and server. It is not a setting that should always be disabled.

kakisoft: Sorry, my phrasing was bad.
I will rephrase it.

## Before:
```
In SSL mode, add "?sslmode=disable"
```

## After:
```
When you want to disable SSL mode, add "?sslmode=disable"
```

k1LoW:
Thank you!

Issue #306, opened 2021-05-11
Doesn't work for MSSQL older than 2017


tjdavis3:
### What happened

`mssql: The function 'STRING_AGG' may not have a WITHIN GROUP clause.`

### What you expected to happened

Document the database

### What stack trace or error message from tbls did you see?

mssql: The function 'STRING_AGG' may not have a WITHIN GROUP clause.

### Anything else we need to know?

STRING_AGG() first appeared in SQL Server 2017 so the use of that function prevents inspection of any but the latest versions of MSSQL.

### Environment

- `tbls version`  1.50.0
- Database version 
       Microsoft SQL Server 2012 (SP2-CU16) (KB3205054) - 11.0.5678.0 (X64) 
	Dec 16 2016 15:17:21 
	Copyright (c) Microsoft Corporation
	Enterprise Edition: Core-based Licensing (64-bit) on Windows NT 6.1 <X64> (Build 7601: Service Pack 1) (Hypervisor)

- OS ( `Ubuntu bionic`, `macOS Calalina`, `WSL2`, etc ) macOS Calalina
- Shell ( `bash`, `zsh`, `poworshell`, etc ) bash
- Stacktrace or error message from tbls ( If you can, set `DEBUG=1`)
  ```
$ DEBUG=1 tbls doc
mssql: The function 'STRING_AGG' may not have a WITHIN GROUP clause.
github.com/k1LoW/tbls/drivers/mssql.(*Mssql).Analyze
	/Users/k1low/src/github.com/k1LoW/tbls/drivers/mssql/mssql.go:175
github.com/k1LoW/tbls/datasource.Analyze
	/Users/k1low/src/github.com/k1LoW/tbls/datasource/datasource.go:113
github.com/k1LoW/tbls/cmd.glob..func5
	/Users/k1low/src/github.com/k1LoW/tbls/cmd/doc.go:71
github.com/spf13/cobra.(*Command).execute
	/Users/k1low/go/pkg/mod/github.com/spf13/cobra@v1.1.1/command.go:850
github.com/spf13/cobra.(*Command).ExecuteC
	/Users/k1low/go/pkg/mod/github.com/spf13/cobra@v1.1.1/command.go:958
github.com/spf13/cobra.(*Command).Execute
	/Users/k1low/go/pkg/mod/github.com/spf13/cobra@v1.1.1/command.go:895
github.com/k1LoW/tbls/cmd.Execute
[...]

PR #310, opened 2021-05-13
fix: Allow mssql driver to work with version older than 2017

tjdavis3:
Changes the SQL statements to remove the STRING_AGG statement.
This change should still work with versions of MSSQL > 2017.

Fixes: #306

Signed-off-by: Todd Davis <tjdavis3@gmail.com>

### DIFF
--- a/drivers/mssql/mssql.go
+++ b/drivers/mssql/mssql.go
@@ -156,21 +156,32 @@ ORDER BY c.column_id
 		constraints := []*schema.Constraint{}
 		/// key constraints
 		keyRows, err := m.db.Query(`
-SELECT
-  c.name,
-  i.type_desc,
-  i.is_unique,
-  i.is_primary_key,
-  i.is_unique_constraint,
-  STRING_AGG(COL_NAME(ic.object_id, ic.column_id), ', ') WITHIN GROUP ( ORDER BY ic.key_ordinal ),
-  c.is_system_named
-FROM sys.key_constraints AS c
-LEFT JOIN sys.indexes AS i ON i.object_id = c.parent_object_id AND i.index_id = c.unique_index_id
-INNER JOIN sys.index_columns AS ic
-ON i.object_id = ic.object_id AND i.index_id = ic.index_id
-WHERE i.object_id = object_id($1)
-GROUP BY c.name, i.index_id, i.type_desc, i.is_unique, i.is_primary_key, i.is_unique_constraint, c.is_system_named
-ORDER BY i.index_id
+		SELECT
+		c.name,
+		i.type_desc,
+		i.is_unique,
+		i.is_primary_key,
+		i.is_unique_constraint,
+		STUFF((
+			Select ', ' + COL_NAME(i.object_id, column_id)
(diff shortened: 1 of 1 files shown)


k1LoW:
Hi @tjdavis3! Thank you for your commit!
Fixed a broken test for the `master` branch. Could you merge latest `master` branch?

drivers/mssql/mssql.go
```diff
@@ -414,7 +454,7 @@ ORDER BY i.index_id
 		}
 		r.Table = table
 		for _, c := range l.columns {
-			column, err := table.FindColumnByName(c)
+			column, err := table.FindColumnByName(strings.TrimSpace(c))
```
k1LoW, line 457: I think it would be better to trim at https://github.com/k1LoW/tbls/pull/310/files#diff-1c6a9273e8cd55f36cf9e47485d049deeb0e1a634afc8a9ebd3970b842ce2bf6R296 or [Query](https://github.com/k1LoW/tbls/pull/310/files#diff-1c6a9273e8cd55f36cf9e47485d049deeb0e1a634afc8a9ebd3970b842ce2bf6R232-R264) than here.

k1LoW (changes requested):
Looking at the test results, it seems that there are spaces in the document that are not needed.
https://github.com/k1LoW/tbls/runs/2580533049?check_suite_focus=true#step:8:842

Could you fix them?

drivers/mssql/mssql.go
tjdavis3: It appears to me that my code changes may not be doing the primary key correctly.  Would you agree based on the test results in lines 807-812?

```
| Name | Type | Definition |
 | ---- | ---- | ---------- |
-| PK__users_* | PRIMARY KEY | CLUSTERED, unique, part of a PRIMARY KEY constraint, [ id ] |
-| UQ__users_* | UNIQUE | NONCLUSTERED, unique, part of a UNIQUE constraint, [ email ] |
-| UQ__users_* | UNIQUE | NONCLUSTERED, unique, part of a UNIQUE constraint, [ username ] |
+| PK__users_* | PRIMARY KEY | CLUSTERED, unique, part of a PRIMARY KEY constraint, [  id, username, email ] |
+| UQ__users_* | UNIQUE | NONCLUSTERED, unique, part of a UNIQUE constraint, [  id, username, email ] |
+| UQ__users_* | UNIQUE | NONCLUSTERED, unique, part of a UNIQUE constraint, [  id, username, email ] |
 | CK__users__username_* | CHECK | CHECK(len([username])>(4)) |
```

k1LoW: > It appears to me that my code changes may not be doing the primary key correctly.

It certainly seems that neither the primary key field nor the unique constraint field is being retrieved correctly.
( mssql DDL is [here](https://github.com/k1LoW/tbls/blob/master/testdata/ddl/mssql.sql) )

There seem to be two problems, 

1. [spaces](https://github.com/k1LoW/tbls/pull/310#pullrequestreview-660299312) 
2. incorrect field retrieval.

PR #329, opened 2021-12-13
Update available commands for tbls

omuomugin:
As I was reading README.md, I realized that message which printed by `tbls help doc` was old and missing some command flags so just adding those for future readers.

However thanks for the fantastic tool.
This is saving our project so just keep up the great work.


k1LoW:
Thank you!

PR #331, opened 2021-12-16
Fix the order of table columns including columnLabels

kymmt90:
Follow up of #328.

Labels introduced by `columnLabels` are placed at the wrong column (the `Children` column) in the output documents. This PR fixes the order of those columns.


k1LoW:
@kymmt90 Thank you !!!!!

PR #332, opened 2021-12-24
Fix typo on go install command

HMasataka:


k1LoW:
@HMasataka Thank you !

PR #338, opened 2022-03-17
Spaces in MD links

YauhenPylAurea:
Thank you for the great tool! 
----

Small fix for md links
image
After applying the fix:
image

----

### DIFF
--- a/output/md/md.go
+++ b/output/md/md.go
@@ -462,7 +462,7 @@ func (m *Md) makeSchemaTemplateData(s *schema.Schema) map[string]interface{} {
 			comment = output.ShowOnlyFirstParagraph(comment)
 		}
 		data := []string{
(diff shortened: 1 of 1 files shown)


output/md/md.go
```diff
@@ -730,3 +730,8 @@ func (m *Md) addNumberToTable(data [][]string) [][]string {
 
 	return data
 }
+
+// replace space by %20 in markdown links (https://spec.commonmark.org/0.30/#link-destination)
+func normalizeMdLink(name string) string {
+	return strings.Replace(name, " ", "%20", -1)
```
k1LoW, line 736: @YauhenPylAurea Thank you for your commit!

Looking at https://spec.commonmark.org/0.30/#link-destination, it looks like there is a little more implementation needed.

Is it possible to add a few more implementations and tests?

k1LoW (changes requested):

PR #340, opened 2022-03-22
Mdlink encoding

YauhenPylAurea:
Continuation of https://github.com/k1LoW/tbls/pull/338

There are 2 issues in the PR that an advise is needed :) Could you please help me with that? Thank you in advance for the help.

1. Encode func is used from [mdurl](https://github.com/golang-commonmark/mdurl). It is archived repository without activity. And it seems only [encode](https://github.com/golang-commonmark/mdurl/blob/master/encode.go) functionality is required to fix the original issue. So the code was copy-pasted. Would it be better to import the whole `mdurl` package instead of the selected approach?

2. Existing test [TestOutput](https://github.com/k1LoW/tbls/blob/main/output/md/md_test.go#L43) has been updated to check the case with a space in a table name. Would it be better to create a completely separate test for such case? 

Thank you.


k1LoW:
> Would it be better to import the whole mdurl package instead of the selected approach?

Yes, tbls uses gocredits to generate CREDITS. I think it is a good idea to use mdurl as a package to follow that mechanism.
It is also meant to honor that package.

Also, it looks like mdurl is moving to the following repository.

https://gitlab.com/golang-commonmark/mdurl

> Would it be better to create a completely separate test for such case?

More test cases are welcome. However, I believe the tests you have modified and added this time are enough.

YauhenPylAurea:
@k1LoW thank you for the review.

Copy-pasted code has been replaced by using `mdurl` package directly as per your advise. Thanks!

k1LoW:
@YauhenPylAurea GREAT!!! Thank you!!!!

k1LoW:
Please wait a little longer for the release!

PR #341, opened 2022-03-31
Stored procedure/functions support for MsSQL, MySQL and PostrgeSQL

YauhenPylAurea:
New section in README.md to show store procedures or user-defined functions, e.g.

## Subroutines

| Name | ReturnType | Arguments | Type |
| ---- | ------- | ------- | ---- |
| dbo.get_user |  | @userid int | SQL inline table-valued function |
| dbo.What_DB_is_that |  | @ID int | SQL Stored Procedure |


Makefile
```diff
@@ -28,7 +28,6 @@ db: db_sqlite
 	usql pg://postgres:pgpass@localhost:55432/testdb?sslmode=disable -f testdata/ddl/postgres95.sql
 	usql pg://postgres:pgpass@localhost:55413/testdb?sslmode=disable -f testdata/ddl/postgres.sql
 	usql my://root:mypass@localhost:33306/testdb -f testdata/ddl/mysql56.sql
-	usql my://root:mypass@localhost:33308/testdb -f testdata/ddl/mysql.sql
```
YauhenPylAurea, line 31: Not sure how to execute such script with [stored procedure and function](https://github.com/k1LoW/tbls/pull/341/files#diff-17bc6cdcfcdef7fa8e80f0062cbccdda0b66f2fbcfa8234f0712934fc143b0f1R108) via `usql`. So it is moved to be executed during [container creation](https://github.com/k1LoW/tbls/pull/341/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R40)

k1LoW: Does this mean that CREATE PROCEDURE could not be executed using `usql` ?

drivers/mssql/mssql.go
```diff
@@ -467,9 +536,17 @@ func (m *Mssql) Info() (*schema.Driver, error) {
 		return nil, err
 	}
 
+	dct := dict.New()
+	dct.Merge(map[string]string{
+		"Subroutines": "Stored procedures and functions",
```
k1LoW, line 541: I am wondering if `Subroutines` is an appropriate name for the section.

k1LoW (commented):
@YauhenPylAurea GREAT!!!!!!! Thank you!!!

I made a few comments, most of them LTGM.

Let me think about it some more 🙏 

Thank you!!!

Makefile
YauhenPylAurea: I wasn't able to execute this file with the following code using `usql`
```
DELIMITER //
CREATE PROCEDURE GetAllComments()
BEGIN
	SELECT * FROM comments;
END//
DELIMITER ;
```

drivers/mssql/mssql.go
YauhenPylAurea: good point. Should I rename it to `ProceduresAndUDF` or something like that?

drivers/mysql/mysql.go
```diff
@@ -487,9 +544,17 @@ func (m *Mysql) Info() (*schema.Driver, error) {
 		name = "mariadb"
 	}
 
+	dct := dict.New()
+	dct.Merge(map[string]string{
+		"Subroutines": "Stored procedures and functions",
```
k1LoW, line 549:
```suggestion
		"Functions": "Stored procedures and functions",
```

I've given it a lot of thought.
How about I proceed with the following naming?

Because, the word "Function" is considered to cover the word "Procedure", and the new word "Subroutine" is not used. I tried not to use it.

k1LoW: I would be happy to have the other "Subroutine" changed as well!

testdata/json_test_schema.json.golden
```diff
@@ -67,6 +67,7 @@
       "virtual": false
     }
   ],
+  "subroutines": null,
```
k1LoW, line 70: Compared to https://github.com/k1LoW/tbls/pull/341/files#diff-7dde8826bb86aa9a1addcc1aedcda0cb76b0b57e2d20d5415cf8fd5ac6baa3b6R49, there seems to be a mixture of null and [], do you know why?

drivers/mssql/mssql.go
```diff
@@ -459,6 +466,68 @@ ORDER BY i.index_id
 	return nil
 }
 
+const query = `select schema_name(obj.schema_id) as schema_name,
```
k1LoW, line 469: If you have the extra power, it would be great if you could change the reserved words in the SQL to uppercase.

drivers/mysql/mysql.go
YauhenPylAurea: fixed in [569f959](https://github.com/k1LoW/tbls/pull/341/commits/569f959e2155bd0747c27959c4b7bcf4587daa4e) and [55d046a](https://github.com/k1LoW/tbls/pull/341/commits/55d046a52e4ec53c8dc8230c822bd51e0a14828a). Two commits as I have some issues with my WSL (my primary OS is windows, so using WSL to build the product on linux) and there are some issues with it. Thanks!

testdata/json_test_schema.json.golden
YauhenPylAurea: fixed in [f0a2289](https://github.com/k1LoW/tbls/pull/341/commits/f0a2289b0c6398cb58c6e3f9e09b057abbc14cea). Is it ok to have empty array?

drivers/mssql/mssql.go
YauhenPylAurea: fixed in [e47d996](https://github.com/k1LoW/tbls/pull/341/commits/e47d996d6407e7ef66f61af8b8468cba327f921f), [faa445f](https://github.com/k1LoW/tbls/pull/341/commits/faa445f5a7c00445a23cf3855ad164a52f8ae750) and [588ced1](https://github.com/k1LoW/tbls/pull/341/commits/588ced15fb2a3a1f817eed1bfb2f7fb15dfca45e)

YauhenPylAurea:
updated per the review. Thank you very much to take a look at this PR!

k1LoW (approved):
GREAT!!!!!!!! Thank you!!!!!!! 🎸

Issue #321, opened 2021-10-21
How to render Columns table without children & parents when there is no information


hey-jude:
I am satisfied with tbls as a simple data discovery tool, and there are some regrets while using it.

Looking at the table columns, there is a children / parents item which is rarely used in my usage. So, if you do not render when there is no children / parents information on the table, I expect to improve the readability of the document.
<img width="1239" alt="Screenshot" src="https://user-images.githubusercontent.com/6929275/138216139-32525b99-1b01-455d-a96a-c925f78c10e0.png">

PR #346, opened 2022-04-06
Hide `parent`, `children`, `comment` columns without data

YauhenPylAurea:
Fix for https://github.com/k1LoW/tbls/issues/321

thank you for the great tool!

it feels like it is possible to make reports more readable by removing columns with no data.


k1LoW:
Thanks for the great pull request.
My current opinions on the additional features are

1. We are concerned that the table formatting will change on a table-by-table basis, which will increase the cognitive load.
    - I expect that the presence or absence of relations and comments will change at a different frequency than Extra Definition in MySQL.
2. I believe documentation is important. Therefore, I would like to be able to tell by an appearance that comments have not been written. (I also think foreign key constraints are a bit important)

I think it is a great feature.
Just that my opinion at this point is the above.

k1LoW:
Maybe my opinion could change in the future.

awhitford-cip:
> 1. We are concerned that the table formatting will change on a table-by-table basis, which will increase the cognitive load.

Snowflake doesn't have parent/child metadata, so that could be a database/connection setting (not unique per table).

YauhenPylAurea:
perhaps we can do it as an option. what do you thing @k1LoW ?

k1LoW:
@YauhenPylAurea Great!!! I had not thought of it at all.

I wanted to add a new section to .tbls.yml.

YauhenPylAurea:
@k1LoW I am not sure what is the best format for such thing in `.tbls.yml`. Should it be under `format`?
```
# .tbls.yml
format:
  # Adjust the column width of Markdown format table
  # Default is false
  adjust: true
  # Sort the order of table list and columns
  # Default is false
  sort: false
  # Display sequential numbers in table rows
  # Default is false
  number: false
  # The comments for each table in the Tables section of the index page will display the text up to the first double newline (first paragraph).
  # Default is false
  showOnlyFirstParagraph: true
  # Hide columns without data
  hideColumnsWithoutData: ["Comment","Parents","Children"]
```
???

k1LoW:
```yaml
format:
  hideColumnsWithoutData: ["Comment","Parents","Children"]
```

I think it is very good.

Now we just need to decide whether to also allow bool values, as follows

```yaml
format:
  hideColumnsWithoutData: true
```

YauhenPylAurea:
What does it mean in such case?
```
format:
  hideColumnsWithoutData: true
```
Any column without data should be hidden, right?

k1LoW:
In the case of an array ( `hideColumnsWithoutData: ["Comment","Parents","Children"]` ), it allows specifying the columns to be hidden.
If `true` ( `hideColumnsWithoutData: true` ), all columns supported to be hidden will be hidden.

The reason why I want to support boolean values is that tbls can change the column names and I think the user will not remember the column names.

k1LoW:
How about I handle the implementation of https://github.com/k1LoW/tbls/pull/346#issuecomment-1116945602 ?

k1LoW (approved):

k1LoW:
@YauhenPylAurea It took a while, but the great features you added have been released as v1.56.0. THANK YOU!!!

PR #350, opened 2022-04-21
doc: Add explanation for hide_auto_increment option

tk0miya:
It was added in #277. But it've never described in the README.  This
adds an explanation for `hide_auto_increment` option.

Thank you for creating such a wonderful tool :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -404,6 +404,13 @@ dsn: mysql://dbuser:dbpass@hostname:3306/dbname
 dsn: my://dbuser:dbpass@hostname:3306/dbname
 ```
 
+When you want to hide AUTO_INCREMENT clause on the table definitions,
+add "?hide_auto_increment".
+For example:
+``` yaml
+dsn: my://dbuser:dbpass@hostname:3306/dbname?hide_auto_increment
+```
+
 **MariaDB:**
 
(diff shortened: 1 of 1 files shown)


tk0miya:
Personally, it would be better to change this option enabled by default. I believe nobody will not get any information from `AUTO_INCREMENT=[Redacted by tbls]`. So I prefer to hide it by default and show it via option.

If you like to change this behavior, it would be better not to merge this. Because it's still a secret option.

k1LoW:
Thank you for adding the documentation!

> Personally, it would be better to change this option enabled by default.

Thank you for your feedback.

I think it would be better to have a default state that first shows the fact that tbls is redacting AUTO_INCREMENT information.
Then, those who want to hide it can enable the `hide_auto_increment`  option, and those who want to show it can enable the `show_auto_increment` option.

k1LoW:
So I want to merge this PR. how do you think about it?

tk0miya:
I'm not sure how AUTO_INCREMENT information is useful. So I'd like to vote to change the default behavior. But I'm follow your decision because we've already had this option :-)

k1LoW:
Thank you !!!

PR #364, opened 2022-07-29
Added sorting to the Table list.

### LINKED ISSUE #362: Table list needs to be sorted
https://github.com/k1LoW/tbls/blob/cf7ec6a772e1d239034ab5de01912e317c72b36b/drivers/snowflake/snowflake.go#L28

I am noticing that TBLS is generating different output, even when there were no database changes.  Specifically, tables are appearing out of order.  I think this is because there is no `order by` clause when querying `information_schema.tables`.

awhitford:
Added sorting to the Snowflake Table list so that the list is consistent.
Fixes #362.

### DIFF
--- a/drivers/snowflake/snowflake.go
+++ b/drivers/snowflake/snowflake.go
@@ -25,7 +25,7 @@ func (sf *Snowflake) Analyze(s *schema.Schema) error {
 	}
 	s.Driver = d
 
-	tableRows, err := sf.db.Query(`SELECT table_name, table_type, comment FROM information_schema.tables WHERE table_schema = ?`, s.Name)
+	tableRows, err := sf.db.Query(`SELECT table_name, table_type, comment FROM information_schema.tables WHERE table_schema = ? order by table_name`, s.Name)
 	if err != nil {
 		return errors.WithStack(err)
 	}


drivers/snowflake/snowflake.go
```diff
@@ -25,7 +25,7 @@ func (sf *Snowflake) Analyze(s *schema.Schema) error {
 	}
 	s.Driver = d
 
-	tableRows, err := sf.db.Query(`SELECT table_name, table_type, comment FROM information_schema.tables WHERE table_schema = ?`, s.Name)
+	tableRows, err := sf.db.Query(`SELECT table_name, table_type, comment FROM information_schema.tables WHERE table_schema = ? order by table_name`, s.Name)
```
k1LoW, line 28: Is it possible to have the tables in order of creation instead of in order of table name?

Is there a column in information_schema.tables to achieve the above in order of creation?

k1LoW (commented):
@awhitford Thank you for your commit!!

I have one question.

drivers/snowflake/snowflake.go
wate: @k1LoW 
MySQL has a creation date column, but it doesn't seem to exist in PostgreSQL

https://dev.mysql.com/doc/refman/8.0/ja/information-schema-tables-table.html
https://www.postgresql.jp/document/13/html/infoschema-tables.html

k1LoW: @wate 
Thank you.
However, the target of this pull request seems to be Snowflake.

wate: @k1LoW 

I believe Snowflake allows sorting by creation date on the `created` column
https://docs.snowflake.com/ja/sql-reference/info-schema/tables.html

k1LoW: @wate Wow!! Thank you!!!

awhitford-cip: > Is it possible to have the tables in order of creation instead of in order of table name?

Why would you want to do that?  I desire a consistent listing of tables in alphabetical order.

k1LoW: If it is in the order of creation, there is a choice to make them in the order intended by the database creator.

> I desire a consistent listing of tables in alphabetical order.

Maybe setting [`format.sort:`](https://github.com/k1LoW/tbls#document-format
) will give you the display you are looking for.

k1LoW: @awhitford-cip @wate 
If sorting by created is selected, the actual order of table creation is unique. However, I found that depending on the speed of creation, created may be at the same time, and the display order may not be unique.

So, I reconsidered that the only way to sort is by table name

awhitford-cip: Right now, my experience is that the table ordering is basically random because it is not defined.  So running tbls against the same database with no material database changes can result in different results, which makes it very difficult to find the table you are interested in when the list is random.

Looking forward to this fix.

PR #365, opened 2022-08-15
Fix extra `-e` displayed on temporary installation.

ogumaru:
Fixed ``-e You can use `tbls` command in this session.`` is displayed after temporary installation finished.
(extra `-e` is displayed.)

How to reproduction.
```bash
wget -O tbls "https://raw.githubusercontent.com/k1LoW/tbls/main/use"
chmod u+x ./tbls
./tbls
```

`-e` option of `echo` is not POSIX, so the option is not recognized when shebang is `#!bin/sh`.

Thank you for updating this nice tool.


k1LoW:
@ogumaru Thank you !!!!

PR #366, opened 2022-08-24
[BREAKING][MariaDB] Change talbe order

k1LoW:


k1LoW:
MariaDB is unordered in information_schema.tables.create_time

Issue #358, opened 2022-06-24
Cross schema foreign keys cannot be procesed


adrianrussell:
### What happened

I have a database with some *odd* to say the least cross schema relationships.  When I run the tool  it halts processing.

I have reproduced with the following sample sql 

```SQL 

USE master;
GO
IF DB_ID (N'db2') IS NOT NULL
DROP DATABASE db2;
GO
CREATE DATABASE db2;
GO

USE db2;
GO

CREATE SCHEMA Sales
GO
CREATE SCHEMA Rabbits
GO

CREATE TABLE [dbo].[Employee] (
    EmployeeID INT PRIMARY KEY CLUSTERED
);
GO
CREATE TABLE [Sales].[Product] (
    ProductID INT PRIMARY KEY CLUSTERED,
	SalesPersonID INT UNIQUE NONCLUSTERED
);
GO

CREATE TABLE [Rabbits].[Running] (
    LocationID INT PRIMARY KEY CLUSTERED,
	ProductID INT UNIQUE NONCLUSTERED,
	EmployeeID INT UNIQUE NONCLUSTERED
);
GO 

--Relationship to a table in another schema causes reporting of the table not being found.
ALTER TABLE [Rabbits].[Running]
   ADD CONSTRAINT FK_TempSales_SalesReason FOREIGN KEY (ProductID)
   REFERENCES [Sales].[Product] (SalesPersonID)
```

### What you expected to happened

I expected that the processing would complete and the relationships between the tables in the different schemas would have been drawn.

### What stack trace or error message from tbls did you see?

Error message reported

`not found table 'Product'`

### Environment

- `tbls version 1.56.0`
- Database version  SQL Server 2019 [Docker Repository](2019-latest
docker pull mcr.microsoft.com/mssql/server:2019-latest)
- OS ( `Docker` )
- Shell (`powershell`)
- Stacktrace or error message from tbls ( If you can, set `DEBUG=1`) `not found table 'Product'`
- Config  `Basic Config`

PR #383, opened 2022-11-07
Cross schema foreign keys cannot be procesed

ypyl:
[Issue](https://github.com/k1LoW/tbls/issues/358)

Not able to run the full set of tests within dev container (e.g. it is not clear how to run dynamodb related tests, also there are some fails with `test_ext_subcommand` from makefile and I really can't understand what it is doing and why it is required.

I will be happy to fix any issues and follow ups from you.
Thank you once again for the great utility.


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,32 @@
+{
```
k1LoW, line 1: I don't think this file needs to be committed.

k1LoW:
@ypyl Thank you for your contribution!!!

.devcontainer/devcontainer.json
RDBModel: fixed

ypyl: fixed in https://github.com/k1LoW/tbls/pull/383/commits/10e897eda3d5f9ff9abebcad264d899bbb6decd9

PR #402, opened 2023-01-05
[BREAKING] Exclude all tables except specific table

### LINKED ISSUE #369: Exclude all tables except specific table
I have a database where I want to exclude all tables from the `auth` schema, except `auth.users`. This database also has tables in other schemas that I want to be included.

I tried
```yaml
# .tbls.yml
include:
  - * # also `"*"`, `'*'`, `"**/*"` as well as removing this line
  - auth.users
exclude:
  - auth.*
```

However, this causes all output to be excluded, producing an empty output.
1. How can I exclude all tables in the `auth` schema, except `auth.users`?
2. How can I include all other tables when using an `include` section? (`*` does not seem to work when used alone)

### Environment

- `tbls 1.56.2`
- OS `Debian bullseye`
- Shell `bash`

k1LoW:
Fix #369 

@FelixZY Thank you!!

### DIFF
--- a/README.md
+++ b/README.md
@@ -767,6 +767,14 @@ lintExclude:
   - CamelizeTable
 ```
 
+#### Filter logic
+
+1. Add tables from include
+2. Remove tables from exclude
+    - Check for include/exclude overlaps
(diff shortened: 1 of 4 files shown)


FelixZY:
@k1LoW thank _you_! Really looking forward to clearing a lot of clutter with this 😃

PR #426, opened 2023-01-27
Fixed position of a comment in the `use` script

rsky:
Following errors were caused due to a comment after `\`. I've fixed it.

```
curl: (3) URL using bad/illegal format or missing URL
curl: (6) Could not resolve host: Skip
curl: (6) Could not resolve host: downloading
curl: (6) Could not resolve host: if
curl: (6) Could not resolve host: the
curl: (6) Could not resolve host: newest
curl: (6) Could not resolve host: file
curl: (6) Could not resolve host: was
curl: (6) Could not resolve host: downloaded
curl: (6) Could not resolve host: before
```

### DIFF
--- a/use
+++ b/use
@@ -33,7 +33,8 @@ download() {
     echo "Unable to get tbls version." >&2
     exit 1
   }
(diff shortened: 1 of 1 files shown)


k1LoW:
@rsky Thank you!

PR #428, opened 2023-02-01
Bulk get schema info for MySQL

yasu89:
# About

In databases with a large number of tables, it took a long time to execute the process, so I improved it.

# Description

The queries to retrieve schema information were issued for each table, resulting in a large overhead and time consuming query processing.

Therefore, we reduced the number of queries by retrieving schema information in batches whenever possible.

## FYI

Result of time in my environment. (over 1,000 tables)

before: about 90min
after: about 3min

### DIFF
--- a/drivers/mysql/mysql.go
+++ b/drivers/mysql/mysql.go
@@ -4,6 +4,7 @@ import (
 	"database/sql"
 	"fmt"
 	"regexp"
+	"sort"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


k1LoW:
@yasu89 GREAT!!!!!!

yasu89:
@k1LoW 
Sorry for the test failure 🙇‍♂️ 

The cause was that the order of the `relations` variable had changed before and after the modification, so I sorted it so that it is the same as before the modification.

PR #430, opened 2023-02-04
Add `er.hideColumnTypes:notRelated:` for hiding not related columns

yasu89:
# About

In a huge database or a table with many columns, the ER diagram becomes large and difficult to look at, so we implemented an option to hide columns without relationships.

# Description

The `FindShowColumnsForER()` method is now used to retrieve a list of columns from a template file via the `FindShowColumnsForER()` method.
When options are enabled in this method, only columns with relationships are returned. 
I have also included this method to make it easier to implement future implementations such as hiding only specific column names.(like `created`, `updated` etc...)


k1LoW:
Thanks for the very nice commitments!!!

> I have also included this method to make it easier to implement future implementations such as hiding only specific column names.(like created, updated etc...)

I think it is a good idea 👍 
How about designing an option that foresees this idea?

For example (It’s my first thought)

```yaml
er:
  typeOfColumnsToShow:
    related: true
    # primary: true  
    # withIndex: false
    # timestamp: false
```

k1LoW:
If you agree with this idea, I can take over the implementation.

yasu89:
@k1LoW 

I thought the optional redesign was a good idea!!

However, in the example you described, setting the columns to show to true does not seem to be able to determine when to show all columns and when to show only the columns that have relationships, so I implemented it by setting the columns to hide to true.

If you have any better ideas for implementation, please point them out and I'll try to implement them.
(or you can take over the implementation if you can)

k1LoW:
Thank you very much. Give me a few moments to think about the interface.

k1LoW:
I implemented `er.showColumnTypes:`. 

#432 

1. Default is to show all.
2. `er.showColumnTypes:` is specified, all columns are hidden by default.
3. `er.showColumnTypes:` must be `true` at least one.
    - `er.showColumnTypes.related:` is always required to be true because currently there is only `related:`.

yasu89:
@k1LoW 

Thank you for the improvements!
(I thought it was very much better than my implementation 🙇‍♂️ )

I have reviewed the PR, not related to the implementation, but just a few things I was concerned about, if you would like to take a look at it.

PR #436, opened 2023-02-08
Add `format.tableGroups` option for table grouping

### LINKED ISSUE #311: Generate schema per Database schema
Love the tool so far!
My whole database is very large, and `schema.svg` file created by **tbls** is very large too, so it's hard to look at it.
I'd love to have an option that would allow me to generate seperate `schema.svg` for every **Database** schema - like `schema-db1.svg`, `schema-db2.svg`.  Splitting the whole database into schemas would be really helpful here.

yasu89:
I am sorry to put out a series of PRs and I hope you will take a look at them when you have time 🙇‍♂️ 

# About
For huge databases, the list of tables in the generated `README.md` is difficult to understand, so I implemented an option to group tables by user-specified groups.

resolve #311 

# Description
In the process of generating `README.md` with `makeSchemaTemplateData()` in `output/md/md.go`, tables are grouped in groups specified by the user and passed to the template.

Also, in the template, when a group of tables is provided, the template is implemented to display a list of tables by group.

In addition, I tried to output ER diagrams between tables within each group.
(In `schema.go`, we implemented a method called `NewSchemaForTableGroup()` that generates a new `Schema` when a list of tables is passed to it, which is then passed to the process that generates the ER diagram.)

# Consultation matter

- How to write a configuration file
  - At first, I thought of writing it as follows, but decided against it because the order is not guaranteed in a Map type, so it would have to be sorted in lexicographic order, and users would not be able to specify the order.
  - If there is a better way, I will re-implement it there.
```yaml
format:
  tableGroups:
    user:
      - users
      - user_options
```

- `README.md` format
  - I thought it would be easier to see if the table groups were closed, so I implemented them as if they were closed using the tags `details` and `summary`.
  - Also, since I implemented the display format according to my own preferences, I would like to re-implement it if there is a better format for this project.


k1LoW:
@yasu89 Thank you for the great feature.

Currently, I am trying to implement a feature in tbls that is similar to this pull request.

It is to introduce the viewpoint/view concept (ref: https://www.viewpoints-and-perspectives.info/ / https://www.sbcr.jp/product/4797376722/ ).

I have already implemented this in another tool called [ndiag](https://github.com/k1LoW/ndiag) ( [ref](https://twitter.com/k1LoW/status/1518762040298733569?s=20&t=vWtsb82tT5HGBIgQEzwt6A) ).

I now consider the following.

1. README.md is not changed. If necessary, have it adjusted using [`templates:`](https://github.com/k1LoW/tbls#personalized-templates) .
2. Create another page according to the `viewpoints:` settings. The page should display each viewpoint's grouped table list, ER diagram, and description.
3. Create links to each viewpoint pages in the README.md if and only if `viewpoints:` is set.

The reason I want to introduce the concept of viewpoints/views is that the grouped tables should have a very important meaning and I want to highlight it.

---

So, your pull request is very nice, but I can't merge it as is. Sorry.

yasu89:
@k1LoW 
Thank you for explaining the future implementation.

I agree that grouped tables are certainly important and should be highlighted.
I look forward to waiting for the new features!

Thank you also for considering my PR!

PR #445, opened 2023-02-20
Fix unclosed fenced code block in README

onk:
"Automatically detect relations" is a very nice feature, but it's not Heading, so I can't link to it.


k1LoW:
@onk Thank you!!

PR #466, opened 2023-04-07
Add a NamingStrategy to map Post.user_id → User.user_id style detected relation

yu-ichiro:
I encountered a niche case where primary keys are named `tablename_id`, and wanted tbls to be able to detect the relation.

This PR adds:
* `identical` and `identicalSingularTableName` NamingStrategy which detects `user_id` → `user_id` relations
* checks to prevent making relation to the table itself

My concern about this is that other NamingStrategy's can be introduced and the combination will blow up. If that's the case, separating the NamingStrategy config into two parts and letting ParentTableNamer and ParentColumnNamer be specified separately might be good maybe??

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -723,6 +723,7 @@ func mergeDetectedRelations(s *schema.Schema, strategy *NamingStrategy) {
 	var (
 		err          error
 		parentColumn *schema.Column
+		parentTable  *schema.Table
 	)
 
 	for _, t := range s.Tables {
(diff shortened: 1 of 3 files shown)


k1LoW:
@yu-ichiro Thank you for your commit!

> My concern about this is that other NamingStrategy's can be introduced and the combination will blow up. If that's the case, separating the NamingStrategy config into two parts and letting ParentTableNamer and ParentColumnNamer be specified separately might be good maybe??

We don't think the number of strategies will increase that much, so we don't think it will be a problem 👍

k1LoW:
Maybe 😄

PR #479, opened 2023-06-06
Fix for when Postgres search path is empty

codetheweb:
Before this change, tbls was throwing for us with `sql: Scan error on column index 0, name "current_schema": converting NULL to string is unsupported`.

This is a similar fix to https://github.com/golang-migrate/migrate/pull/914.


k1LoW:
@codetheweb Thank you !!!!!

PR #510, opened 2023-09-30
Add an arm64 architecture variant of the Linux build

mmizutani:
This PR adds an arm64 CPU architecture variant of the Linux build as a release target in addition to the existing amd64 (x86_64) Linux build.

Currently, unlike [`tbls-build`](https://github.com/k1LoW/homebrew-tap/blob/9d4fbacd69cfdc4fd3cd780f45c9e6de92d4ddeb/tbls-build.rb#L17-L26), the release builds of the `tbls` program for macOS are available for both the amd64 and arm64 architectures to support both Intel CPU and Apple CPU, but the release builds for Linux OSes are only available for the amd64 architecture.

There are a couple of use cases where the arm64 arch Linux release of the `tbls` program is valuable:

- Installing `tbls` in a Linux OS container on an Apple CPU macOS machines without resorting to compatibility mode (enabling emulation of the linux/amd64 platform), which could hurt performance and development productivity.
- ~~Installing `tbls` on Raspberry Pi devices, which are powered by ARM processors.~~ This is armhf arch, not arm64.
- Installing `tbls` in workloads (e.g., self-hosted CI/CD pipelines) running on virtual machines and cloud container platforms powered by the [AWS Graviton CPU](https://aws.amazon.com/ec2/graviton/), which is based on the ARM CPU architecture.

PR #512, opened 2023-10-02
Fix linux arm64 release build

mmizutani:
The Linux arm64 arch build requires installing an additional OS package `gcc-aarch64-linux-gnu`. I have altered the CI test (dry-run) config file to install this package in the PR #510, but I forgot to also update the CI config file for real (non-dry-run) releases.

This PR resolves the [error](https://github.com/k1LoW/tbls/actions/runs/6386314568/job/17332776671) during the last release build for Linux caused by this omission.

> cgo: C compiler "aarch64-linux-gnu-gcc" not found: exec: "aarch64-linux-gnu-gcc": executable file not found in $PATH

### DIFF
--- a/.github/workflows/tagpr.yml
+++ b/.github/workflows/tagpr.yml
@@ -58,7 +58,7 @@ jobs:
       - name: Install packages
         run: |
           sudo apt-get update
-          sudo apt-get install sqlite3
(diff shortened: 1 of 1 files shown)


k1LoW:
oh...Thank you!

Issue #525, opened 2023-10-16
Add "Arguments" to the field to sort in function category.


corydoraspanda:
### What happened

When creating functions with the same name but different numbers of arguments, sorting might not work as expected.

For example, in postgres, if you repeatedly create and delete the following function, the difference will appear with `tbls diff`.

```sql
CREATE OR REPLACE FUNCTION add(integer, integer) RETURNS integer
    AS 'select $1 + $2;'
    LANGUAGE SQL
    IMMUTABLE
    RETURNS NULL ON NULL INPUT;

CREATE OR REPLACE FUNCTION add(integer, integer, integer) RETURNS integer
    AS 'select $1 + $2 + $3;'
    LANGUAGE SQL
    IMMUTABLE
    RETURNS NULL ON NULL INPUT;
```

diff

```diff
root@aa87a023bf4b:/workspace/tbls# ./tbls diff pg://postgres:pgpass@postgres95/testdb?sslmode=disable  --sort -c testdata/test_tbls_postgres.yml sample/postgres95a
diff 'sample/postgres95a/README.md' 'tbls doc pg://postgres:*****@postgres95/testdb?sslmode=disable'
--- sample/postgres95a/README.md
+++ tbls doc pg://postgres:*****@postgres95/testdb?sslmode=disable
@@ -29,8 +29,8 @@

 | Name | ReturnType | Arguments | Type |
 | ---- | ------- | ------- | ---- |
+| public.add | int4 | integer, integer | FUNCTION |
 | public.add | int4 | integer, integer, integer | FUNCTION |
-| public.add | int4 | integer, integer | FUNCTION |
 | public.update_updated | trigger |  | FUNCTION |
 | public.uuid_generate_v1 | uuid |  | FUNCTION |
 | public.uuid_generate_v1mc | uuid |  | FUNCTION |
```

### What you expected to happened

Function arguments are also included in the sorting target.

### What stack trace or error message from tbls did you see?

Nothing

### Anything else we need to know?

Proposal for how to fix it

```diff
diff --git a/schema/schema.go b/schema/schema.go
index 2b9f6f0..6097e76 100644
--- a/schema/schema.go
+++ b/schema/schema.go
@@ -290,7 +290,10 @@ func (s *Schema) Sort() error {
 		return s.Relations[i].Table.Name < s.Relations[j].Table.Name
 	})
 	sort.SliceStable(s.Functions, func(i, j int) bool {
[...]

PR #527, opened 2023-10-17
Add arguments as second sort key for functions.

corydoraspanda:
request for #525.

I added arguments as the second sort key in functions.
I've also added test code.


k1LoW:
@corydoraspanda GREAT!!!!! Thank you!!

Issue #526, opened 2023-10-17
Refer viewpoints from table of Markdown


macoto1995:
Hi. I'm currently writing viewpoints for most tables with tbls.
It is working great to describe the data structure.

Now I'm getting to want each table of markdown to have reference to viewpoints 
Then, we can tell how each table was created and aggregated in context of multiple tables.

## Indexes
...

## Viewpoints

| Name | Definition |
| ---- | ---------- |
| [(name of the viewpoint)](/#link-to-viewpoint-md) | (description of the viewpoint) |

## Relations
...

k1LoW:
@macoto1995 Nice idea!!!! LGTM!

PR #530, opened 2023-10-22
feat:refer viewpoints from tables

macoto1995:
## Description
Impleemented #526 

## Type of change
- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

updated golden test files

### DIFF
--- a/output/output.go
+++ b/output/output.go
@@ -15,7 +15,7 @@ import (
 // Output is interface for output
 type Output interface {
 	OutputSchema(wr io.Writer, s *schema.Schema) error
-	OutputTable(wr io.Writer, s *schema.Table) error
+	OutputTable(wr io.Writer, s *schema.Table, viewpoints schema.Viewpoints) error
 }
 
 var escapeMermaidRe = regexp.MustCompile(`[^a-zA-Z0-9_\-]`)
--- a/output/config/config.go
+++ b/output/config/config.go
@@ -95,6 +95,6 @@ func (c *Config) OutputSchema(wr io.Writer, s *schema.Schema) error {
 	return nil
 }
 
-func (c *Config) OutputTable(wr io.Writer, t *schema.Table) error {
+func (c *Config) OutputTable(wr io.Writer, t *schema.Table, _v schema.Viewpoints) error {
 	return errors.New("not supported")
 }
--- a/output/dot/dot.go
+++ b/output/dot/dot.go
@@ -58,7 +58,7 @@ func (d *Dot) OutputSchema(wr io.Writer, s *schema.Schema) error {
 }
 
 // OutputTable output dot format for table.
-func (d *Dot) OutputTable(wr io.Writer, t *schema.Table) error {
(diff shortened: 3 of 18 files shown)


output/output.go
```diff
@@ -15,7 +15,7 @@ import (
 // Output is interface for output
 type Output interface {
 	OutputSchema(wr io.Writer, s *schema.Schema) error
-	OutputTable(wr io.Writer, s *schema.Table) error
+	OutputTable(wr io.Writer, s *schema.Table, viewpoints schema.Viewpoints) error
```
macoto1995, line 18: @k1LoW 
To refer viewpoints from tables, I need to chenge the `Output` interface and that affects other unrelated files like `dot.go`.
If you have better idea to implement this feature, kindly let me know.
I appletiate it.

macoto1995:
TODO: run all tests to update golden files.

output/output.go
k1LoW: It may be better to add a [Viewpoints](https://github.com/k1LoW/tbls/blob/5059fd987a596def1d1c350da019745a0a84d6fe/schema/schema.go#L72) field to the [Table](https://github.com/k1LoW/tbls/blob/5059fd987a596def1d1c350da019745a0a84d6fe/schema/schema.go#L135-L147).

k1LoW: @macoto1995 You may have it implemented as is. I will correct it on my end!

macoto1995: @k1LoW 
Oh..Sorry I missed your new comment for a few days.

> It may be better to add a [Viewpoints](https://github.com/k1LoW/tbls/blob/5059fd987a596def1d1c350da019745a0a84d6fe/schema/schema.go#L72) field to the [Table](https://github.com/k1LoW/tbls/blob/5059fd987a596def1d1c350da019745a0a84d6fe/schema/schema.go#L135-L147).
I have some concern about this solution. 
Since `Viewpoint` has `Tables` as children, it would be infinite loop.
Plus, to render link to viewpoints from the markdown, the `OutputTable` function has to know the index of the viewpoints, which is not included in `Viewpoint` type.

So I suppose that it's better to have `TableViewpoint` type like below.

```go
struct TableViewpoint {
    Index Int
    Name String
    Desc String
}
```

Kindly, let me know what you think.
I'd work on this way for now. 
Thank you.

k1LoW: > Since Viewpoint has Tables as children, it would be infinite loop.

Yes. In fact, recursive relationships are everywhere (e.g. `Table.ReferencedTables` and `Table.Column.ParentRelations` ).

> Plus, to render link to viewpoints from the markdown, the OutputTable function has to know the index of the viewpoints, which is not included in Viewpoint type.

fmfm, indeed. Good point.

> So I suppose that it's better to have TableViewpoint type like below.

LGTM. So far I have no better ideas either.

macoto1995: Thanks! I'll work on this way!

macoto1995:
moved to #532

PR #572, opened 2024-04-13
Update go-github-client version and and use new options in `NewGithubClient`.

kromiii:
I want to add GitHub App Authentication in `AnalyzeGitHubContent` function. 

* Upgrade `go-github-client` from `v58.0.11` to `v58.0.12`, by which we can use `factory.OwnerRepo` function.
* By using `factory.OwnerRepo` function, we can pass the option to `NewGithubClient()`. It will enable us to use GitHub App authentication in `AnalyzeGitHubContent` function.

My old PR (https://github.com/k1LoW/tbls/pull/571) failed CI, so I arranged it, and re-created the PR.


k1LoW:
@kromiii Thank you!

PR #581, opened 2024-05-06
Use `ghfs` for AnalyzeGithubContent

kromiii:
## Problem
Sometimes, fetching a schema file from GitHub fails because the file exceeds the file size limit in GitHub API.

https://docs.github.com/ja/rest/repos/contents?apiVersion=2022-11-28#get-contents

```
Failed to analyze schema: unsupported content encoding: none, this may occur when file size > 1 MB, if that is the case consider using DownloadContents
```

## Solution
We can avoid this by using `ghfs` (https://github.com/k1LoW/ghfs). It uses Git Database API, which has no file size limit.

## Concern
I would appriciate if you would judge whether we need to rewrite the entire code to accommodate some large files.


datasource/datasource.go
```diff
@@ -167,25 +166,22 @@ func AnalyzeGitHubContent(dsn config.DSN) (*schema.Schema, error) {
...
-	if f == nil {
-		return nil, errors.Errorf("invalid dsn: %s", dsn)
-	}
-	cc, err := f.GetContent()
+	b, err := fsys.ReadFile(splitted[2])
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
-	dec := json.NewDecoder(strings.NewReader(cc))
+	dec := json.NewDecoder(strings.NewReader(string(b)))
```
k1LoW, line 184:
```suggestion
	dec := json.NewDecoder(bytes.NewReader(b))
```

k1LoW (changes requested):
GREAT! Thank you!!

One minor correction, please.

kromiii:
> GREAT! Thank you!!
> 
> One minor correction, please.

Thanks!

k1LoW (approved):
LGTM!

PR #594, opened 2024-06-19
Postgres: Support "$user" search path

mjpieters:
This simply replaces the string `"$user"` in the Postgres search path with the current username.


k1LoW:
@mjpieters GREAT!! Thank you!

PR #595, opened 2024-06-19
Address lint errors

mjpieters:
- Remove unused config functions (these were moved to `filter.go` in 1bed38d)
- Use entries of mongodb bson.D result directly instead of using the deprecated Map method.
- Don't ignore err results from utility functions.

I really like being able to rely on `make lint` being clean, seeing errors there, even if not in my own PR changes, makes me itch. 😅 

Specifically, these lint issues have been addressed:

```
golangci-lint run ./...
drivers/mysql/mysql.go:501:49: ineffectual assignment to err (ineffassign)
                strColumns, strParentTable, strParentColumns, err := parseFK(r.Def)
                                                              ^
drivers/mongodb/mongodb.go:98:27: SA1019: result.Map is deprecated: Converting directly from a D to an M will not be supported in Go Driver 2.0. Instead, users should marshal the D to BSON using bson.Marshal and unmarshal it to M using bson.Unmarshal. (staticcheck)
                for key, value := range result.Map() {
                                        ^
drivers/sqlite/sqlite.go:368:49: ineffectual assignment to err (ineffassign)
                strColumns, strParentTable, strParentColumns, err := parseFK(r.Def)
                                                              ^
config/config.go:553:6: func `excludeTableFromSchema` is unused (unused)
func excludeTableFromSchema(name string, s *schema.Schema) error {
     ^
config/config.go:908:6: func `matchLabels` is unused (unused)
func matchLabels(il []string, l schema.Labels) bool {
     ^
```

### DIFF
--- a/drivers/mongodb/mongodb.go
+++ b/drivers/mongodb/mongodb.go
@@ -95,7 +95,8 @@ func (d *Mongodb) listFields(collection *mongo.Collection) ([]*schema.Column, er
 			return columns, err
 		}
 		total += 1
-		for key, value := range result.Map() {
+		for _, entry := range result {
+			key, value := entry.Key, entry.Value
 			var valueType string
 			switch value.(type) {
 			case string:
--- a/config/config.go
+++ b/config/config.go
@@ -550,47 +550,6 @@ func (c *Config) mergeDictFromSchema(s *schema.Schema) {
 	}
 }
 
-func excludeTableFromSchema(name string, s *schema.Schema) error {
-	// Tables
-	tables := []*schema.Table{}
-	for _, t := range s.Tables {
-		if t.Name != name {
-			tables = append(tables, t)
-		}
-		for _, c := range t.Columns {
-			// ChildRelations
-			childRelations := []*schema.Relation{}
-			for _, r := range c.ChildRelations {
-				if r.Table.Name != name && r.ParentTable.Name != name {
-					childRelations = append(childRelations, r)
-				}
-			}
-			c.ChildRelations = childRelations
-
-			// ParentRelations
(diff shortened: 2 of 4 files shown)


mjpieters:
I suspect that these lint issues have been missed for some time because you set the `golanglint` action to use the `github-pr-check` reporter. These end up as error annotations on the workflow summary page (see my [other PR workflow summary](https://github.com/k1LoW/tbls/actions/runs/9580508207?pr=594)), and the lint check itself **doesn't show an error**.

I recommend you change the configuration for that action to make these more visible:

- You could enable [`fail_on_error`](https://github.com/reviewdog/action-golangci-lint?tab=readme-ov-file#fail_on_error) to have the action fail the workflow when there are errors.
- You could try and see if the `reporter: github-pr-review` option is more visible.

drivers/mongodb/mongodb.go
```diff
@@ -95,7 +95,8 @@ func (d *Mongodb) listFields(collection *mongo.Collection) ([]*schema.Column, er
 			return columns, err
 		}
 		total += 1
-		for key, value := range result.Map() {
+		for _, entry := range result {
+			key, value := entry.Key, entry.Value
```
mjpieters, line 99: Note: I chose this option over unmarshalling as a `bson.M` map to maintain the order preservation.

**However** by using `result.Map()` the order was already undone, and before this change the key-value pairs were iterated over in hash table order.

If this is now resulting in unexpected ordering changes, we could change the above to use `var result bson.M` instead and use `for key, value := range result {` for this loop.

k1LoW:
@mjpieters Thank you!! I am very pleased with these improvements.

PR #600, opened 2024-06-21
Refactor when expression handling

### LINKED ISSUE #599: when expression parsing too simplistic
tbls runs `when` expressions with the [`expr` package](https://expr-lang.org/), but first replaces any `$ENV_VAR` string with the the string, if defined in the environment, with `Env.ENV_VAR`.

There are two problems with this approach:

- If the expression uses the special [`$expr` variable](https://expr-lang.org/docs/language-definition#env), _and_ the environment variable `env` exists, then the special variable is replaced by `Env.env`.
- The text replacement can trivially replace text in string literals or other parts of the expression, even though they are not identifiers. E.g. `env FOO="bar" BAR='$FOO' tbls --when '$BAR == "$FOO"'` should run (`BAR` is equal to the literal string `$FOO`), but fails because the string literal `"$FOO"` in the expression was replaced by `"Env.FOO"`.

The solution is to use [`ast.Patch()`](https://expr-lang.org/docs/patch) after compiling the expression into an AST to walk, and replacing each `$REF` name with `Env.REF` (as a `MemberNode` with Env identifier and `REF` string literal). This can be done for **all** such names (excepting `$env`) so `Env.NONESUCHNAME` properly fails as a `unknown field NONESUCHNAME` error.

I'll create a PR.

mjpieters:
- Use expr.Patch and a visitor to only replace identifiers that start with `$` with a `Env.` member lookup.
- use `expr.AsBool()` to assert the expression produces a boolean
- Clean up envMap parsing to use strings.Cut()
- Expand tests, using testing.T.Run() with test names.

Fixes #599

### DIFF
--- a/cmdutil/when.go
+++ b/cmdutil/when.go
@@ -1,49 +1,62 @@
 package cmdutil
 
 import (
-	"fmt"
 	"os"
 	"strings"
 
 	"github.com/expr-lang/expr"
+	"github.com/expr-lang/expr/ast"
 	"github.com/pkg/errors"
 )
+
+// AST walker which replaces `$IDENTIFIER` with `Env.IDENTIFIER` member lookup expressions.
+type EnvPatcher struct{}
+
+func (EnvPatcher) Visit(node *ast.Node) {
+	if n, ok := (*node).(*ast.IdentifierNode); ok && n.Value[0] == '$' && n.Value != "$env" {
+		ast.Patch(
+			node,
+			&ast.MemberNode{
+				Node:     &ast.IdentifierNode{Value: "Env"},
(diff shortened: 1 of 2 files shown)


mjpieters:
Note that with this update, using undefined variables in an expression now provides a much friendlier error:

```
./tbls --when 'foo'        
unknown name foo (1:1)
 | foo
 | ^
 ```
 
Before you got:

```
./tbls --when 'foo'
cannot fetch foo from struct { Env map[string]string } (1:2)
 | (foo) == true
 | .^
 ```

k1LoW:
@mjpieters Looks Great!!!  thank you!!!

PR #601, opened 2024-06-25
Add support to output enum definitions in the db doc README

insano10:
I'd like enum support for the documentation.
It's not finished yet as there's no tests, but can someone give me feedback if I'm going down the right lines?
I'm only using Postgres at the moment which is why it's only been implemented for that driver.

When I build and run tbls against my local db, the enum output looks as expected in the README.md.


insano10:
To be perfectly honest I was waiting for the maintainer to give an indication that if I worked on this, I would be given feedback and it would likely be merged. I don't have a lot of free time right now and didn't want to spend time fixing up tests only for it to sit there.

There does seem to be some desire for it though so if no one else wants to add it, I'll try pick it up soon.

k1LoW:
@insano10 Thank you for your contribution!

I think it is a very good idea.

Let me give you just two pieces of feedback.

1. if the Enum does not exist, it would be better not to mention anything in the documentation
2. some tests seem to fail, many of them may succeed by fixing 1.

insano10:
@k1LoW 
I believe I've got all the tests passing now. 
I haven't been able to test all the different types of DBs, only against my local Postgres but it works well for that (for databases with/without enums).

drivers/postgres/postgres.go
```diff
@@ -490,6 +497,31 @@ func (p *Postgres) getFunctionsByQuery(query string) ([]*schema.Function, error)
...
 }
 
+func (p *Postgres) getEnums() ([]*schema.Enum, error) {
+	enums := []*schema.Enum{}
+
+	enumsResult, err := p.db.Query(`SELECT t.typname AS enum_name, ARRAY_AGG(e.enumlabel) AS enum_values
+										  FROM pg_type t, pg_enum e
+										  WHERE t.typcategory = 'E'
+										  AND t.oid = e.enumtypid
+										  GROUP BY t.typname `)
```
k1LoW, line 507: It doesn't seem to be narrowed down by the schema.

k1LoW:
Please check the test results as well, as we have fixed the CI environment.

insano10:
@k1LoW Thank you for your feedback. I have fixed the test failures and updated the enum definitions to include their schema.

drivers/postgres/postgres.go
```diff
@@ -490,6 +497,41 @@ func (p *Postgres) getFunctionsByQuery(query string) ([]*schema.Function, error)
...
+			enumName   string
+			enumValues []string
+		)
+		err := enumsResult.Scan(&schemaName, &enumName, pq.Array(&enumValues))
+		if err != nil {
+			return enums, errors.WithStack(err)
+		}
+
+		enum := &schema.Enum{
+			Name:   fmt.Sprintf("%s.%s", schemaName, enumName),
```
k1LoW, line 527: 👍 NICE 👍

insano10:
@k1LoW Apologies I forgot to update the test expectation after adding the schema scoping. Fixed now.

k1LoW:
@insano10 Thank you!!!! GREAT!

insano10:
@k1LoW Thanks so much for shepherding this through. The `setup-tbls` github action has now picked up this release so my CI build output now includes enums 🙌

PR #605, opened 2024-07-06
Add support for ClickHouse

joschi:
Closes #156

PR #608, opened 2024-07-23
Add ClickHouse support to README

joschi:
Add note about ClickHouse support to README.md.

This was missing in #605 before.


k1LoW:
@joschi Thank you!

PR #621, opened 2024-10-01
Fix CI

k1LoW:
SSIA.


k1LoW:
📝 

```
tbls-mssql-1        mcr.microsoft.com/mssql/server:2017-latest   "/opt/mssql/bin/nonr…"   mssql        22 seconds ago   Restarting (1) Less than a second ago   
```

https://github.com/k1LoW/tbls/actions/runs/11121830826/job/30901804241#step:9:1731

PR #634, opened 2024-11-12
Fix schema filter logic to prevent duplicated counts of tables

kromiii:
When I tested the schema filter function with a Distance of 3 in a very small schema for the development, I noticed the following error often occurs in filter.go:

> Failed to load schema: failed to filter schema: failed to separate tables. expected: 10, actual: 16

By checking the function, I found a bug in the aggregation logic of the filter function, which caused duplicate tables to be counted in `includes2`, so I had fix it.


k1LoW:
@kromiii Thank you!

PR #644, opened 2025-01-20
touch up grammar, punctuation, and Markdown

brandondrew:
I haven't gone through the whole README, but I found a number of things that could be improved. 

Although "for" was sometimes used like this in archaic forms of English (Middle English?  Early Modern?) it is not used in current Modern English (and I think it was always "for to" back when it _was_ used—I don't remember).

In regard to code fences, it does appear that GitHub allows the language identifier to have a space before it (and after the 3 backticks) but their guide shows it without a space, and I'm not sure whether other Markdown parsers would allow that space (which I've never seen before).  If you really like the space I can modify the PR to add them back, but I think it's much safer to do without them.

https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks#syntax-highlighting


k1LoW:
@brandondrew Thank you!

PR #654, opened 2025-02-01
Fix errcheck related to type assertion

bmf-san:
# Description
I noticed that errcheck was returning an error by running `make lint`.
It seems to be caused by missing return value checks in a type assertion.

```sh
Error return value is not checked (errcheck)
```

# Steps to Reproduce
```sh
make lint
```

# What I Did
I addressed the root cause of the errcheck issue by adding conditional branching to the type assertion.

I decided to keep the function signature the same.

Although ignoring the error could be an option to avoid redundant error handling, I decided to handle it properly rather than ignoring it carelessly.

# What I confirmed
- The error disappeared with `make lint`
- The test cases for the changed parts passed

# Others
This is my first PR to tbls.
I often use tbls because they are easy to use and useful.
Thanks for tbls. :D

### DIFF
--- a/cmdutil/when.go
+++ b/cmdutil/when.go
@@ -1,6 +1,7 @@
 package cmdutil
 
 import (
+	"fmt"
 	"os"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


k1LoW:
@bmf-san Thank you!!!

PR #671, opened 2025-03-13
Express tbls.yml comments override schema comments

tnagatomi:
Currently, the behavior where comments in `tbls.yml` override schema comments is not explicitly documented or tested.

So I added documentation to the README, and modified the test for the overriding case.


k1LoW:
@tnagatomi Thank you!

PR #718, opened 2025-07-01
fix: Skip enum fetching for Redshift compatibility in postgres driver

ritsu2891:
Amazon Redshift does not support PostgreSQL enum types:  
https://docs.aws.amazon.com/redshift/latest/dg/c_unsupported-postgresql-datatypes.html

When running `tbls` against Redshift, the enum type query causes an error.  
> ERROR: relation "pg_enum" does not exist

This PR skips fetching enum types when the driver is Redshift, allowing documentation generation to proceed without failure.


k1LoW:
@ritsu2891 Thank you!!!! LGTM!

PR #727, opened 2025-07-21
[WIP] Support for custom template paths

reteps:
This PR adds 4 features:

- Output path templating for the markdown output files
- Output path templating for graphviz output files
- Support for an `Enum` template in markdown
- Support to skip file generation by setting the output path to the empty string

This code with the help of Claude 4.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -1038,9 +1038,23 @@ templates:
   md:
     index: 'templates/index.md.tmpl'
     table: 'templates/table.md.tmpl'
+    enum: 'templates/enum.md.tmpl'
 ```
 
 A good starting point to design your own template is to modify a copy the default ones for [Dot](output/dot/templates), [PlantUML](output/plantuml/templates) and [markdown](output/md/templates).
+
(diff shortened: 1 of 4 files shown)


k1LoW:
Thanks for the contribution!

I don't understand why this fix is needed, and there are many fixes in one. I cannot accept it at this time.

reteps:
Hi @k1LoW -- if I split each change up into it's own PR, would that help each feature be accepted? These features are super useful for customizing the output of tbls for our project.

reteps:
To provide some more context, I am trying to replace our in-house solution with tbls, and these are the minimum features to add to replace it. The path templating is useful because it gives us extra flexibility in creating directories/files in our preferred structure.

k1LoW:
It would be great to know in what cases each modification would be useful and why it is necessary (and why it needs to be incorporated as a feature in tbls).

It may be less costly to have the conversation in an Issue rather than splitting this Pull Request first.

PR #737, opened 2025-08-14
 Add databricks support

Ewan-Keith:
firstly, thanks for the tool! 

This PR adds support for databricks sql based on othe ther implementations. DBx does have support for primary/foreign keys, although they're not enforced. As they're still extremely useful for documentation I've pulled them through via the driver.

Some unit testing is included but as a cloud platform databricks doesn't make integration testing hugely easy. I have copied the snowflake TPC-H SF1 DDL in `./testdata` to use DBx SQL syntax and have used this to carry out manual testing. The output of this testing is stored at `sample/databricks`. Databricks has recently released a free version which can be spun up with just an email (no payment details or cloud infra required) so if the maintainers wanted to it should be relatively straightforward to setup e2e testing.

2 auth mechanisms are supported, depending on whether a user or service principal is being used to execute the commands. This means that the DSN Ive landed on for tbls use isn't exactly the same as the internal dsn used by the dbx go library but this felt like an acceptable tradeoff for  a clearer interface.

finally, while local testing I wanted to avoid accidentally committing my own dbx creds so have added `tbls.yml` to the .gitignore file. I don't know if that's desirable or not, feel free to remove it if not.

I _think_ I've got everywhere that needs updating for a new database updated, but just point me to any I've missed!


README.md
```diff
@@ -607,6 +607,32 @@ dsn: clickhouse://dbuser:dbpass@hostname:9000/dbname
 
 See also: https://pkg.go.dev/github.com/ClickHouse/clickhouse-go
 
+**Databricks:**
+
+**Personal Access Token (PAT) Authentication:**
```
k1LoW, line 612: To ensure consistency with other data sources, I would appreciate it if you could either discontinue the use of bold formatting or add explanatory comments within the code blocks.

README.md
```diff
@@ -607,6 +607,32 @@ dsn: clickhouse://dbuser:dbpass@hostname:9000/dbname
 
 See also: https://pkg.go.dev/github.com/ClickHouse/clickhouse-go
 
+**Databricks:**
```
k1LoW, line 610:
```suggestion
**Databricks (Experimental):**
```

As this is a data source I cannot actively maintain and continuous testing via Docker images or similar cannot be performed, please allow me to designate it as ‘Experimental’.

k1LoW (commented):
@Ewan-Keith GREAT WORK!!

Thank you!!

I have two minor requests for revision. Please consider them.

k1LoW (approved):
Looks GREAT!! Thank you!!

PR #749, opened 2025-10-03
feat(drivers): Support foreignKey for spanner

seipan:
Thank you for this great project!
We are using Spanner in our service and wanted to adopt this library. Since it did not support foreign keys, I added support for them.


k1LoW:
@seipan Thank you!!!

PR #758, opened 2025-10-18
Changed HTML line-break output from <br> to self-closing <br /> in adjustTable.

Mistat:
This pull request makes a small formatting update to the Markdown output in the `md.go` file. The change ensures that all line breaks in table data are consistently replaced with the HTML-compliant `<br />` tag instead of `<br>`. This improves compatibility with strict HTML parsers.

- Updated the `strings.NewReplacer` in the `adjustTable` function to use `<br />` for line breaks in table data, ensuring HTML compliance.


k1LoW:
@Mistat Thank you!!

PR #777, opened 2025-12-08
fix(mssql): exclude HEAP indexes from index retrieval query

ejsmile:
Problem:
--------
When analyzing MSSQL tables with NONCLUSTERED PRIMARY KEY constraints, the driver was incorrectly including HEAP index entries in the results. This caused issues because HEAP entries have NULL names and represent the base table storage structure rather than actual indexes.

In SQL Server's sys.indexes catalog view:
- i.type = 0 represents a HEAP (table without clustered index)
- i.type = 1 represents a CLUSTERED index
- i.type = 2 represents a NONCLUSTERED index
- i.type = 3-7 represent specialized index types (XML, SPATIAL, etc.)

When a table has only a NONCLUSTERED PRIMARY KEY and no CLUSTERED index, SQL Server stores the table data as a HEAP (type = 0, name = NULL). This HEAP entry was being included in query results, causing:
1. Index entries with NULL names
2. Incorrect index counts
3. Potential downstream processing errors

Example query result BEFORE fix:
  name                              index_id  type  type_desc
  --------------------------------  --------  ----  ------------
  NULL                              0         0     HEAP
  PK_tableWithOutClusterIndex       2         2     NONCLUSTERED
  IX_testIndex                      3         2     NONCLUSTERED

Solution:
---------
Added filter condition 'AND i.type <> 0' to the index retrieval query in drivers/mssql/mssql.go. This explicitly excludes HEAP entries while preserving all actual index types (CLUSTERED, NONCLUSTERED, XML, etc.).

Using i.type instead of i.index_id for filtering is semantically clearer:
- i.type <> 0 explicitly means 'exclude HEAP storage structures'
- i.index_id has special meaning (0 for HEAP/XML, 1 for CLUSTERED) but filtering by it would be less obvious to maintainers

Example query result AFTER fix:
  name                              index_id  type  type_desc
  --------------------------------  --------  ----  ------------
  PK_tableWithOutClusterIndex       2         2     NONCLUSTERED
  IX_testIndex                      3         2     NONCLUSTERED

Changes:
--------
1. Added test table 'tableWithOutClusterIndex' in testdata/ddl/mssql.sql with NONCLUSTERED PRIMARY KEY to reproduce the issue
2. Added integration test 'TestTableWithNonClusteredPrimaryKey' to verify:
   - HEAP indexes are excluded from results
   - Correct number of indexes are returned
   - No indexes have empty names
   - Primary key is properly identified
3. Modified index query in drivers/mssql/mssql.go to filter HEAP entries

Tested on:
- Local MSSQL Server with tables having NONCLUSTERED PRIMARY KEY
- Docker container with MSSQL Server 2019
- All integration tests passing

### DIFF
--- a/sample/mssql/name with spaces.md
+++ b/sample/mssql/name with spaces.md
@@ -6,12 +6,10 @@
 <summary><strong>Table Definition</strong></summary>
 
 ```sql
-
-CREATE VIEW [name with spaces] AS (
+CREATE VIEW "name with spaces" AS (
   SELECT TOP 1 p.title
   FROM posts AS p
 );
-
 ```
 
 </details>


k1LoW:
@ejsmile GREAT!!! Thank you!!!

Issue #731, opened 2025-07-24
Empty `Description` section in generated Markdown output


reteps:
### What happened

The `Description` header is present in the default Markdown template, even if there is no description content

### Expected behavior

This header is only present if there is content in the section. This bug is fixed in #727

PR #810, opened 2026-03-21
fix: hide Description header when table has no comment or definition

KeitaShimura:
## Summary

The default Markdown template always rendered a ## Description header even when a table had no comment and no definition. This PR fixes the template to skip the entire Description section when both fields are empty.

## Changes

- **output/md/templates/table.md.tmpl**: Wrap the ## Description section in a conditional so it is only rendered when .Table.Comment or .Table.Def is present.
- **sample/**: Updated all sample markdown files across multiple drivers (postgres95, postgres, mssql, spanner, dynamodb, bigquery_census_bureau_international, adjust) to remove empty ## Description sections, aligning them with the new template output.

## Motivation

An empty ## Description section adds noise to generated documentation and is misleading — it implies there is content when there is none.

## Test Plan

- `go test -v ./output/md/...` — all tests pass, including `TestOutput` (golden file comparison for generated markdown) and `TestDiffSchemaAndDocs`.

## Related

Related to #731

### DIFF
--- a/sample/adjust/backup.blog_options.md
+++ b/sample/adjust/backup.blog_options.md
@@ -1,7 +1,5 @@
 # backup.blog_options
 
-## Description
-
 ## Columns
 
 | Name    | Type                        | Default                                  | Nullable | Children | Parents                         | Comment |
--- a/sample/adjust/backup.blogs.md
+++ b/sample/adjust/backup.blogs.md
@@ -1,7 +1,5 @@
 # backup.blogs
 
-## Description
-
 ## Columns
 
 | Name    | Type                        | Default                           | Nullable | Children                                      | Parents | Comment |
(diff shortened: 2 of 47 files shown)


KeitaShimura:
Updated sample markdown files across multiple drivers (postgres95, postgres, mssql, spanner, dynamodb, bigquery_census_bureau_international, adjust) to remove empty `## Description` sections, aligning them with the new template output.

This was necessary because the CI `testdoc` step performs a diff between the generated output and the sample files — the samples were still containing the old empty `## Description` sections, causing the check to fail.

Verified with `go test -v ./output/md/...` — all tests pass.

- [fix: update all samples to remove empty Description headers](https://github.com/k1LoW/tbls/pull/810/commits/2a566a5bc693f7fe069ee525fa7bc47008828d57)

PR #802, opened 2026-02-03
feat: add --verbose and --skip-partitions flags for doc/out commands

patkujawa-wf:
NOTE: I used Claude Code for these changes.

If this is desirable, I'm happy to split out the `--skip-partitions` changes and the `--verbose` changes into two PRs (or drop the verbose one if not desired).

## Summary

- Add `--verbose`/`-v` flag to show progress with elapsed timestamps during database analysis
- Add `--skip-partitions` flag to skip PostgreSQL table partitions (useful for databases with many partitions)
- Skip relations referencing excluded partition tables when `--skip-partitions` is enabled
- Mask passwords in DSN URLs when verbose logging is enabled
- Add verbose logging throughout datasource connection and postgres driver to help diagnose slow operations

## Motivation

When running `tbls doc` against a PostgreSQL database through a proxy (e.g., Aurora), the command can appear to hang with no feedback. These changes help users:

1. See real-time progress to identify where time is spent
2. Skip partition tables that can significantly slow down analysis (each partition requires separate queries for columns, indexes, constraints, etc.)

## Example output

```
[   0.00s] Starting tbls doc
[   0.00s] Loading configuration
[   0.00s] Configuration loaded (DSN: postgres://user:****@host:5432/db)
[   0.00s] Analyzing database schema
[   0.00s] Opening database connection
[   5.32s] Pinging database
[  10.45s] Database connection established
[  10.45s] Running driver analysis for postgres
[  10.46s] Fetching database version info
[  10.52s] Querying tables
[  10.53s] Skipping table partitions
[  12.30s] Processing table: public.users (BASE TABLE)
...
```

## Test plan

- [x] Verified `--verbose` shows progress with SQLite test database
- [x] Verified `--skip-partitions` excludes partition tables
- [x] Verified password masking in verbose output
- [x] Verified `tbls out` command also supports both flags
- [x] Tested against real Aurora PostgreSQL with partitioned tables

PR #827, opened 2026-04-20
feat: add --verbose flag for doc/out commands

patkujawa-wf:
## Summary

Splits #802 into two focused PRs. This PR adds the `--verbose` / `-v` flag to `tbls doc` and `tbls out`.

- Adds `--verbose/-v` flag printing progress with elapsed timestamps to stderr
- Masks passwords in DSN when verbose logging
- Adds verbose logging to datasource connection and the postgres driver
- Extracts the postgres tables query into a `queryForTables()` helper (pure refactor)

The companion PR adds `--skip-partitions` on top of this one.

## Test plan

- [ ] `tbls doc -v <dsn> <path>` prints timestamped progress lines
- [ ] `tbls out -v -t json <dsn>` prints timestamped progress lines
- [ ] DSN password is masked in the logged output
- [ ] Without `-v`, output is unchanged

### DIFF
--- a/drivers/postgres/postgres.go
+++ b/drivers/postgres/postgres.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/aquasecurity/go-version/pkg/version"
 	"github.com/k1LoW/errors"
+	"github.com/k1LoW/tbls/cmdutil"
 	"github.com/k1LoW/tbls/ddl"
 	"github.com/k1LoW/tbls/dict"
 	"github.com/k1LoW/tbls/schema"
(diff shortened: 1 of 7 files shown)


drivers/postgres/postgres.go
```diff
@@ -98,24 +102,8 @@ func (p *Postgres) Analyze(s *schema.Schema) (err error) {
 	fullTableNames := []string{}
 
 	// tables
-	tableRows, err := p.db.Query(`
```
patkujawa-wf, line 101: Scope creep so there's clearer/fewer changes in https://github.com/k1LoW/tbls/pull/828/changes#diff-e36a277bcc301a326320a2ca3b839ee745128b68008e5200c67313804898758fR745-R760 but can back this out if desired

k1LoW (commented):
Thank you for the PR!

I appreciate the effort, but I don't currently see a need for a `--verbose` flag in tbls. I understand that schema analysis can be slow depending on the database, but I believe that's the only part that can take significant time. Verbose logging across the entire command doesn't seem necessary for that.

patkujawa-wf:
Fair enough. I added this in order to understand why scanning our postgres db was taking so long and it revealed to me the behavior of including all partitioned tables, so I found it valuable at least :)

PR #830, opened 2026-04-27
fix(mysql): stabilize constraint introspection order

### LINKED ISSUE #829: Make MySQL constraint output deterministic
Currently, MySQL constraint output can be non-deterministic when a table has multiple CHECK constraints.

The order of CHECK constraints may change between runs for the same schema, which causes unnecessary diffs in generated documentation.

For example, rows in the constraints table can be reordered between runs:

```diff
-| chk_users_email_valid | CHECK | ...
 | chk_users_status_valid | CHECK | ...
+| chk_users_email_valid | CHECK | ...
```

The MySQL driver appends CHECK constraints in the order returned by the `information_schema` query, but the query currently has no `ORDER BY`.

A minimal improvement would be to add an explicit order to the CHECK constraint query:

```sql
ORDER BY t.table_name, c.constraint_name
```

It may also be worth stabilizing the top-level order of PK/UNIQUE/FK constraint rows. Their column order is already stabilized with `GROUP_CONCAT(... ORDER BY ...)`, but the constraint row order itself is not explicitly ordered.

Using `format.sort: true` can avoid this diff, but it also sorts columns by name, so fixing the driver introspection query seems less invasive.

I can send a PR if this direction looks reasonable.

upamune:
## Summary

This PR makes MySQL constraint introspection deterministic by adding explicit `ORDER BY` clauses to constraint queries.

It stabilizes:

- CHECK constraint order by table name and constraint name
- PK/UNIQUE/FK constraint row order by table name, constraint name, constraint type, and referenced table

This prevents unnecessary diffs in generated docs when MySQL returns `information_schema` rows in a different order.

Closes #829

## Testing

```sh
go test ./drivers/mysql
go test -tags mysql ./drivers/mysql
```

### DIFF
--- a/drivers/mysql/mysql.go
+++ b/drivers/mysql/mysql.go
@@ -396,7 +396,8 @@ ON kcu.constraint_name = sub.constraint_name
   AND kcu.table_name = sub.table_name
   AND (kcu.referenced_table_name = sub.referenced_table_name OR (kcu.referenced_table_name IS NULL AND sub.referenced_table_name IS NULL))
 WHERE kcu.table_schema= ?
-GROUP BY kcu.table_name, kcu.constraint_name, sub.constraint_type, kcu.referenced_table_name`, s.Name)
+GROUP BY kcu.table_name, kcu.constraint_name, sub.constraint_type, kcu.referenced_table_name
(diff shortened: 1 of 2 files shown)


k1LoW (approved):
LGTM!

PR #833, opened 2026-05-07
Ensure consistent column ordering for ClickHouse

Leward:
When running `tbls` against a ClickHouse instance, I noticed that sometimes it would randomly change the order of the columns in the rendered Markdown table.

By adding `position` to the `ORDER BY` clause, we ensure that the column list is always returned in the same order.


Leward:
I wanted to add tests, but it's a behavior that is difficult to consistently replicate :/

k1LoW:
@Leward Thank you for your contribution!!!

PR #838, opened 2026-06-06
feat: add Microsoft Fabric Warehouse / Azure SQL support via Service Principal

Germain-D:
- New azuresql:// datasource — connects to Microsoft Fabric Warehouse and Azure SQL using Azure AD
  Service Principal auth (fedauth=ActiveDirectoryServicePrincipal). Delegates schema analysis to the
  existing mssql driver.
  - Replace FOR XML PATH with STRING_AGG — FOR XML PATH is unsupported in Synapse SQL / Microsoft Fabric
  Warehouse. STRING_AGG is supported on Azure SQL, Fabric Warehouse, and SQL Server 2017+.

  Breaking Change

  SQL Server 2016 and earlier no longer supported (no STRING_AGG).

  Files changed

  - datasource/azuresql.go — new datasource + STRING_AGG fix
  - datasource/azuresql_test.go — tests for both
  - datasource/datasource.go — register new datasource
  - drivers/mssql/mssql.go — remove FOR XML PATH
  - go.mod / go.sum — new Azure SDK dependency


k1LoW:
@Germain-D GREAT!! Thank you!!!

I'm unable to maintain a persistent test environment for Microsoft Fabric Warehouse/Azure SQL.

If I put Microsoft Fabric Warehouse/Azure SQL support in tbls, is it ok to make it "experimental support"?

Also, is it possible to commit the following files?

1. a sample DDL similar to other data sources ( in [testdata/](https://github.com/k1LoW/tbls/tree/master/testdata) )
2. a sample document generated by tbls doc for the database created in 1 ( in [sample/](https://github.com/k1LoW/tbls/tree/master/sample) )
3. Datasource section in README.md

datasource/azuresql.go
```diff
@@ -0,0 +1,85 @@
...
+	dbName := u.Query().Get("database")
+	if dbName == "" {
+		return "", fmt.Errorf("no database name in azuresql connection string")
+	}
+	return dbName, nil
+}
+
+// buildAzureSQLConnStr converts azuresql:// URL to a key=value connection string
+// because go-mssqldb/azuread only URL-parses sqlserver:// scheme, not azuresql://.
+func buildAzureSQLConnStr(urlstr string) (string, error) {
```
k1LoW: **Consider rewriting the scheme instead of hand-building a `key=value` connection string**

`go-mssqldb/azuread` registers the `azuresql` driver, but its DSN parser (`msdsn.Parse`) only *URL-parses* the `sqlserver://` scheme. Instead of assembling an ADO `key=value` string by hand, we can keep the URL form and just swap the scheme `azuresql://` → `sqlserver://`, then pass it to `sql.Open("azuresql", ...)`.

I checked `azuread/configuration.go`: for `ActiveDirectoryServicePrincipal` it reads `params["user id"]` (split into tenant/client) and `params["password"]`, and `msdsn` merges + percent-decodes URL query params (`msdsn/conn_str.go`), so the existing DSN shape keeps working unchanged.

Besides removing both `buildAzureSQLConnStr` and `parseAzureSQLDatabaseName`, this fixes a latent issue: the current `fmt.Sprintf` builds the ADO form with no escaping, and `msdsn`'s default (non-`odbc:`) parser splits on `;` with no brace quoting. So a password containing `;` would break parsing or inject extra options (e.g. `TrustServerCertificate=true`). The URL form avoids that, since query values are percent-decoded rather than `;`-split.

Rough sketch:

```go
func AnalyzeAzureSQL(urlstr string) (_ *schema.Schema, err error) {
	defer func() { err = errors.WithStack(err) }()

	u, err := url.Parse(urlstr)
	if err != nil {
		return nil, err
	}
	q := u.Query()
	dbName := q.Get("database")
	if dbName == "" {
		return nil, fmt.Errorf("no database name in azuresql connection string")
	}
	if q.Get("fedauth") == "" {
		q.Set("fedauth", "ActiveDirectoryServicePrincipal")
	}
	q.Set("encrypt", "true") // preserve current Encrypt=true posture
	u.RawQuery = q.Encode()
	u.Scheme = "sqlserver" // azuread driver only URL-parses sqlserver://

	db, err := sql.Open("azuresql", u.String())
	// ... Ping + mssqlDriver.New(db).Analyze(s)
}
```

WDYT?

Germain-D: I think it's a great idea!  I tried implementing it with Claude. Let me know what you think

datasource/azuresql.go
```diff
@@ -0,0 +1,63 @@
...
+		return "", "", fmt.Errorf("no database name in azuresql connection string")
+	}
+	if q.Get("fedauth") == "" {
+		q.Set("fedauth", "ActiveDirectoryServicePrincipal")
+	}
+	q.Set("encrypt", "true")
+	u.RawQuery = q.Encode()
+	u.Scheme = "sqlserver"
+
+	return u.String(), dbName, nil
```
k1LoW, line 48: Proposal: could the Service Principal credentials be carried in the URL **userinfo** instead of the `user id` / `password` query parameters?

```
# ?database= as today
azuresql://<client-id>@<tenant-id>:<client-secret>@host?database=mydb&fedauth=ActiveDirectoryServicePrincipal

# fedauth omitted, falling back to the ActiveDirectoryServicePrincipal default
azuresql://<client-id>@<tenant-id>:<client-secret>@host?database=mydb

# tenant omitted, resolved from the server
azuresql://<client-id>:<client-secret>@host?database=mydb&fedauth=ActiveDirectoryServicePrincipal

# path form, if the datasource.go:120-125 style translation is added
azuresql://<client-id>@<tenant-id>:<client-secret>@host/mydb?fedauth=ActiveDirectoryServicePrincipal
```

Only the credentials move. `database`, `fedauth` and anything else stay where they are.

### Why

1. **The client secret is currently printed in cleartext by `tbls diff`.** `Config.MaskedDSN()` (`config/config.go:601`) masks only the userinfo password and returns the DSN untouched when it is not set. `output/md/md.go:229,339` feed that value into the `tbls doc <dsn>` diff header, so the secret ends up in the diff output and in CI logs, including `make test_azuresql`. Moving it into userinfo makes masking work with no change to `MaskedDSN`.
2. **It matches the shape already documented for SQL Server** just above the new block: `sqlserver://DbUser:SQLServer-DbPassw0rd@hostname:1433/testdb`.
3. **`azuresql` is already a dburl alias for the sqlserver driver** (`xo/dburl` `scheme.go`, aliases `ms`, `mssql`, `azuresql`), and `sqlserver` is in `supportDriversWithDburl`, so `azuresql://user:pass@host/db` reaches the generic path today. The unconditional branch in `datasource.go` turns those into `no database name in azuresql connection string`. The userinfo form stays compatible with what already worked.
4. It removes the `user+id` vs `user id` distinction from the docs, where `+` is only the query-string encoding.

### It works out mechanically

- msdsn maps userinfo straight onto the params azuread reads: `splitConnectionStringURL` sets `res[UserID]` / `res[Password]` from `u.User` (`msdsn/conn_str.go:875-881`), and the ServicePrincipal branch reads `params["user id"]` / `params["password"]` (`azuread/configuration.go:131`).
- **No escaping burden.** Go splits userinfo from host at the *last* `@`, so a raw `@` in the client id parses correctly and round-trips through `u.String()` as `%40`:
  ```
  azuresql://cid@tid:sec@host/mydb    -> user="cid@tid" pass="sec" host="host" path="/mydb"
  azuresql://cid%40tid:sec@host/mydb  -> user="cid@tid" pass="sec" host="host" path="/mydb"
  azuresql://cid:sec@host/mydb        -> user="cid"     pass="sec" host="host" path="/mydb"
  ```
- Tenant stays optional: `splitTenantAndClientID` (`azuread/configuration.go:298-306`) splits on the first `@` and returns `(user, "")` when there is none, in which case the server's tenant is used.
- `?database=` is still needed at the msdsn layer, since it turns `u.Path` into a named instance (`conn_str.go:888-889`). But tbls already does that translation for sqlserver at `datasource/datasource.go:120-125`, so the `/mydb` path form could be accepted with the same few lines if that is wanted.

### Related, separate question

Would `ActiveDirectoryDefault` be a better default than `ActiveDirectoryServicePrincipal`? It is the one mode that needs no secret in the DSN at all, since azidentity picks up `AZURE_CLIENT_ID` / `AZURE_TENANT_ID` / `AZURE_CLIENT_SECRET` or a managed identity. That tends to fit CI and cloud execution better, and it would leave the secret-bearing form as an explicit opt-in.

Happy to be wrong on any of this if there is a constraint I am missing.

Germain-D: Thank you for the thorough review and great suggestions @k1LoW!

I completely agree with this approach. I have updated the implementation accordingly:

1. **Userinfo credentials**: Credentials can now be passed via userinfo (`azuresql://<client-id>@<tenant-id>:<client-secret>@host/mydb`). As you noted, this automatically makes `Config.MaskedDSN()` mask the secret (`*****`) in `tbls diff` and logs without any core changes.
2. **Database path & msdsn instance fix**: Database name can be specified in the URL path (`/mydb`) or via `?database=mydb`. `u.Path` is then cleared before passing to `go-mssqldb` so `msdsn.Parse` does not interpret `/mydb` as a named instance (`host\mydb`).
3. **Smart `fedauth` default**:
   - If credentials are present (userinfo or query), `fedauth` defaults to `ActiveDirectoryServicePrincipal`.
   - If no credentials are provided (`azuresql://host/mydb`), `fedauth` defaults to `ActiveDirectoryDefault` (picking up `AZURE_*` env vars, Azure CLI, or Managed Identity).
   - Any explicit `?fedauth=...` is preserved.
4. **Compatibility**: Query parameters (`?user+id=...&password=...`) and optional tenant form (`<client-id>:<client-secret>`) remain supported.

Tests and documentation (`[README.md](http://readme.md/)`, `Makefile`) have been updated.

k1LoW:
@Germain-D Great work!!! Thank you!!

Issue #786, opened 2025-12-21
Panic on Arch Linux


artemklevtsov:
Hi,

Can't start `tbls` from the gitgub release page on arch.

Compiled from the source code works fine.

```sh
# ./tbls 
SIGFPE: floating-point exception
PC=0x7f4597d697bf m=19 sigcode=1
signal arrived during cgo execution
instruction bytes: 0x49 0xf7 0xf0 0x48 0x29 0xd1 0x48 0x89 0xf2 0x48 0x8d 0x84 0xe 0x0 0x8 0x0

goroutine 39 gp=0xc000603a40 m=19 mp=0xc000581808 [syscall]:
runtime.cgocall(0x301cb40, 0xc0000e1d00)
        /opt/hostedtoolcache/go/1.24.10/x64/src/runtime/cgocall.go:167 +0x4b fp=0xc0000e1cd8 sp=0xc0000e1ca0 pc=0x47358b
github.com/snowflakedb/gosnowflake._Cfunc_dlOpen(0x7f45a4000b70)
        _cgo_gotypes.go:129 +0x45 fp=0xc0000e1d00 sp=0xc0000e1cd8 pc=0x1a73ac5
github.com/snowflakedb/gosnowflake.loadFromPath({0xc001c66730, 0x42})
        /home/runner/go/pkg/mod/github.com/snowflakedb/gosnowflake@v1.18.1/minicore_posix.go:68 +0xad fp=0xc0000e1e18 sp=0xc0000e1d00 pc=0x1a73d6d
github.com/snowflakedb/gosnowflake.(*miniCoreLoaderType).loadCore(0x3d49ab5?)
        /home/runner/go/pkg/mod/github.com/snowflakedb/gosnowflake@v1.18.1/minicore.go:254 +0x2c2 fp=0xc0000e1f90 sp=0xc0000e1e18 pc=0x1a3dea2
github.com/snowflakedb/gosnowflake.getMiniCore.func1.1()
        /home/runner/go/pkg/mod/github.com/snowflakedb/gosnowflake@v1.18.1/minicore.go:302 +0xdf fp=0xc0000e1fe0 sp=0xc0000e1f90 pc=0x1a74cbf
runtime.goexit({})
        /opt/hostedtoolcache/go/1.24.10/x64/src/runtime/asm_amd64.s:1700 +0x1 fp=0xc0000e1fe8 sp=0xc0000e1fe0 pc=0x47e861
[...]

PR #840, opened 2026-06-22
fix: bump gosnowflake to v2.1.0 and disable minicore at build time

k1LoW:
## Summary

Bump `github.com/snowflakedb/gosnowflake` from v1.18.0 to v2.1.0 and build with the `minicore_disabled` tag. This removes the startup `dlopen` of gosnowflake's minicore (native Rust extension) that crashed the statically-linked release binaries, and lets us unfreeze the driver version that #799 pinned as a workaround.

## Changes

- **go.mod / go.sum / main.go / drivers/snowflake/snowflake.go**: Move the driver import to the v2 module path (`github.com/snowflakedb/gosnowflake/v2`). The driver is still a blank import for `database/sql` registration only; the registered driver name (`"snowflake"`) is unchanged, so the connection code needs no changes despite the v2 major bump.
- **.goreleaser/{linux,windows,darwin}.yml**: Add `minicore_disabled` to the build `-tags` so release binaries never compile in minicore.
- **Makefile**: Add `minicore_disabled` to the `build` target so local builds match the release configuration.

## Motivation

#786 reported that the pre-built release binary crashes with `SIGFPE` at startup on Arch Linux, even when Snowflake is never used. The release binaries are statically linked (`-linkmode external -extldflags '-static'`), and on such binaries gosnowflake's minicore loader calls `dlopen`, which crashes against a newer glibc. #799 worked around this by freezing gosnowflake at v1.18.0.

gosnowflake v1.19.0+ fixes this properly: it auto-detects static linking and skips minicore, and exposes a `minicore_disabled` build tag to exclude minicore from the binary entirely. tbls does not use any minicore-backed functionality, so disabling it has no effect on normal use while eliminating the crash and shrinking the binary.

Refs #786


k1LoW:
## Verified the #786 crash is gone on Arch Linux

Reproduced the original report's setup locally: built a static `linux/amd64` binary on a debian/glibc `golang` image (mirroring `.goreleaser/linux.yml`), then ran it on `archlinux:latest` (cutting-edge glibc), which is the environment that triggered the `SIGFPE`.

Build flags used:

- `github.com/snowflakedb/gosnowflake/v2 v2.1.0`
- `CGO_ENABLED=1`, `-tags 'netgo timetzdata sqlite_fts5 minicore_disabled'`
- `-ldflags "-linkmode external -extldflags '-static'"`

Run on `archlinux:latest`:

```console
$ ldd ./tbls
	not a dynamic executable
$ ldd --version | head -1
ldd (GNU libc) 2.43
$ ./tbls            # the #786 repro: no args
Usage:
  tbls [command]
...
# EXIT=0  (no SIGFPE)
$ ./tbls version
1.94.5
# EXIT=0
```

Before (v1.18.0 with minicore enabled) the same `./tbls` invocation crashed with `SIGFPE` during the minicore `dlopen`. With `minicore_disabled` that `dlopen` path is not compiled in, so the crash is gone while the `database/sql` driver registration still works.

PR #841, opened 2026-06-30
fix: escape pipe characters in column cells of Markdown table output

mahirhir:
In the `## Columns` table, the `Comment` and `Extra Definition` cells are written through `mdEscRep` (which escapes `|`, `<` and `>`), but the `Name`, `Type` and `Default` cells on the same row are written raw:

```go
data := []string{
    c.Name,
    c.Type,
    c.Default.String,
    fmt.Sprintf("%v", c.Nullable),
}
adjustData(&data, ..., mdEscRep.Replace(c.ExtraDef))
...
adjustData(&data, ..., mdEscRep.Replace(c.Comment))
```

When one of those raw cells contains a `|` (for example a string or expression default such as `'a|b'`), the pipe is read as a cell separator, the row gains an extra column, and the table loses its alignment.

For a column with `default = 'a|b'` and `comment = "one | two"` the output is:

```
| Name | Type | Default | Nullable | Children | Parents | Comment |
| status | text | 'a|b' | false |  |  | one \| two |
```

The header has 7 cells but the data row has 8: the comment pipe is escaped (`one \| two`) while the default pipe is not, so the data row no longer matches the header.

This applies `mdEscRep.Replace` to the `Name`, `Type` and `Default` cells, the same way `Comment` and `Extra Definition` already are. The existing golden files do not contain these characters, so they are unchanged; the added test covers the pipe case.

### DIFF
--- a/output/md/md.go
+++ b/output/md/md.go
@@ -579,9 +579,9 @@ func (m *Md) makeTableTemplateData(t *schema.Table) map[string]interface{} {
 		}
 
 		data := []string{
-			c.Name,
-			c.Type,
-			c.Default.String,
+			mdEscRep.Replace(c.Name),
+			mdEscRep.Replace(c.Type),
+			mdEscRep.Replace(c.Default.String),
 			fmt.Sprintf("%v", c.Nullable),
 		}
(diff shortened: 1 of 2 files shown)


k1LoW:
@greymoth-jp Thank you!

PR #844, opened 2026-07-06
perf(spanner): fetch schema in bulk to avoid N+1 queries

YuheiTakagawa:
## Summary

The Spanner driver's `Analyze()` queried `INFORMATION_SCHEMA` **per table** (columns, indexes)
and **per column** (column options). This is an N+1 pattern that dominates `tbls doc` /
`tbls out` runtime on non-trivial schemas.

This PR fetches columns, column options, and indexes with **one bulk query each** and groups
the rows in Go, mirroring the pattern the `mysql` driver already uses for columns and the
foreign-key query already used in this same driver. Query count becomes constant (5),
independent of table/column count. **The generated output is unchanged.**

## Problem

For a schema with `T` tables and `C` total columns, `Analyze()` issued:
1 (tables)

T (columns, one query per table)
C (column options, one query per column) ← worst offender
T (indexes, one query per table)
1 (foreign keys, already bulk)
= 2 + 2T + C queries
The per-column `COLUMN_OPTIONS` query is the biggest contributor.

## Benchmark
Measured against the Cloud Spanner emulator with a real-world schema of **135 tables / 977 columns**,
built from source, warm-up run discarded, `tbls out -t json` (median of 5 runs):
| | before (this driver on `main`) | after (this PR) |
|---|---|---|
| queries | **1249** | **5** |
| wall time | **~194 s** | **~1.2 s** |
~160x faster. On a remote Spanner instance the gap is larger, since each of the ~1249
round-trips pays network latency.
## Change
- Columns: single query `... ORDER BY TABLE_NAME, ORDINAL_POSITION`, grouped per table in Go.
- Column options: single query, applied to columns via a `map[table]map[column]*Column`.
- Indexes / primary keys: single query (dropped the per-table `WHERE`, kept the existing
  `GROUP BY`), `ORDER BY TABLE_NAME, INDEX_NAME`.
- Foreign keys: unchanged (already bulk).
`ORDER BY` was added to the bulk queries so column/index ordering is deterministic; the
previous per-table queries had no top-level `ORDER BY`, so ordering was unspecified.
## Compatibility / how verified
The schema output is identical before and after this change:
- `tbls out -t json` before vs after → **byte-identical** on
  (a) `testdata/ddl/spanner.sql` and (b) the 135-table schema above (MD5 match).
- `go build ./...`, `go vet ./drivers/spanner/`, `gofmt` clean.
- The existing `//go:build spanner` integration test (`drivers/spanner/spanner_test.go`)
  compiles and is unaffected.
> Note: benchmarking/verification used the Spanner **emulator**.

### DIFF
--- a/drivers/spanner/spanner.go
+++ b/drivers/spanner/spanner.go
@@ -37,7 +37,7 @@ func (sp *Spanner) Analyze(s *schema.Schema) error {
 	}
 	s.Driver = d
 
-	// tables / constraints
+	// tables
 	tableStmt := spanner.Statement{SQL: `
 SELECT
   TABLE_NAME, PARENT_TABLE_NAME, ON_DELETE_ACTION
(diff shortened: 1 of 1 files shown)


k1LoW:
@YuheiTakagawa Thank you for your contribution!!

PR #847, opened 2026-07-21
feat: support ssl-ca, ssl-cert and ssl-key DSN parameters

mtawil:
## Why

Connecting to databases that use a private CA or require client certificates (e.g. managed MySQL with `REQUIRE X509` accounts) is currently not possible for drivers whose Go SQL driver has no file-based TLS DSN parameters. The go-sql-driver/mysql `tls` parameter only accepts `true|false|skip-verify|preferred`; loading certificate files requires `RegisterTLSConfig()`, which a DSN-only tool like tbls never calls. This makes tbls unusable in CI/Kubernetes environments where the database mandates certificate-based TLS.

## What

Adds four generic DSN query parameters — `ssl-ca`, `ssl-cert`, `ssl-key` and `ssl-verify-identity` — applied through each driver's own TLS mechanism:

| Driver | Behavior |
|---|---|
| `mysql` / `maria` | Loads the files, registers a per-call TLS config via `mysql.RegisterTLSConfig`, and rewrites `tls=` to use it. `ssl-ca` alone verifies the chain without hostname verification (`VERIFY_CA`); adding `ssl-verify-identity=true` also verifies the hostname (`VERIFY_IDENTITY`). An explicit `tls=true` keeps full verification; `tls=false`/`tls=preferred` conflict and are rejected |
| `postgres` / `rs` | Translates to lib/pq's native `sslrootcert` / `sslcert` / `sslkey`; sets `sslmode=verify-ca` when no `sslmode` was given (an explicit verifying `sslmode` is kept — lib/pq treats `require` + root cert like `verify-ca`), and `sslmode=verify-full` for `ssl-verify-identity=true` |
| `sqlserver` | `ssl-ca` maps to go-mssqldb's `certificate` (`.pem`/`.der` file) with `encrypt=true` and `trustservercertificate=false`; an explicit `encrypt=strict` is kept; `ssl-cert`/`ssl-key` return an error — SQL Server has no client-certificate authentication |
| others | Explicit error — including the non-dburl datasources (BigQuery, Spanner, DynamoDB, MongoDB, Databricks), so the parameters are never silently ignored |

Combined with the drivers' existing native parameters (`tls=` for MySQL/MariaDB, `sslmode=` for PostgreSQL, `encrypt=` for SQL Server), every standard SSL mode is now expressible.

Additional semantics:

- Explicit native TLS parameters that contradict the requested verification (`sslmode=disable`, `encrypt=disable`, `trustservercertificate=true`, …) are rejected with an error instead of being silently overridden or ignored.
- `ssl-cert`/`ssl-key` present a client certificate and must be set together. The client certificate is supplied via `GetClientCertificate`, so it is sent even when the server's acceptable-CA list does not match the issuer.
- Chain verification uses `tls.Config.VerifyConnection` (not `VerifyPeerCertificate`), so it also covers resumed sessions and passes gosec's G123 check.
- TLS configs are registered under a unique per-call name, so concurrent library use cannot clobber another call's configuration.
- **Empty values are treated as absent** and stripped from the DSN, so config templates like `?ssl-ca=${MYSQL_SSL_CA}` keep working unchanged when the environment variables are not set.
- The parameters are always removed before the DSN reaches the SQL driver (same pattern as `hide_auto_increment`).

## Docs

- New "SSL/TLS" section under "DSN" in the README documenting the generic parameters and the conflict rules, with driver-specific mapping notes in the PostgreSQL and Microsoft SQL Server sections.
- The MySQL section documents how each `--ssl-mode` of the mysql client maps to DSN parameters.

## Testing

- `golangci-lint run` (the repo config, including gosec) passes with 0 issues.
- Unit tests cover all dispatch paths (mysql/maria, postgres/redshift, sqlserver, unsupported and non-dburl datasources), the chain-verification callback itself (valid chain, wrong CA, missing certificate), conflict rejection for every driver, unique config-name registration, empty-value stripping, `ssl-verify-identity` parsing, cert-without-key validation, and bad/missing/invalid certificate files (test certificates are generated in-test).
- Verified end-to-end against a real MySQL 8.0 server: connects with the full `ssl-ca` + `ssl-cert` + `ssl-key` trio; `tls=skip-verify&ssl-ca=…` upgrades to chain verification; `ssl-verify-identity=true` correctly rejects a server certificate without a matching SAN; `tls=preferred&ssl-ca=…` is rejected as a conflict.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -69,7 +69,7 @@ type Config struct {
 type DSN struct {
 	URL     string            `yaml:"url"`
 	Headers map[string]string `yaml:"headers,omitempty"`
-	TLS     TLS               `yaml:"tls,omitempty"`
+	TLS     *TLS              `yaml:"tls,omitempty"`
 }
 
 // TLS is the SSL/TLS configuration for the datasource connection.
--- a/config/config.go
+++ b/config/config.go
@@ -69,6 +69,15 @@ type Config struct {
 type DSN struct {
 	URL     string            `yaml:"url"`
 	Headers map[string]string `yaml:"headers,omitempty"`
+	TLS     TLS               `yaml:"tls,omitempty"`
+}
+
+// TLS is the SSL/TLS configuration for the datasource connection.
+type TLS struct {
+	CA     string `yaml:"ca,omitempty"`
+	Cert   string `yaml:"cert,omitempty"`
+	Key    string `yaml:"key,omitempty"`
+	Verify string `yaml:"verify,omitempty"`
 }
 
 // Format is document format setting.
--- a/datasource/datasource.go
+++ b/datasource/datasource.go
@@ -56,6 +56,14 @@ func Analyze(dsn config.DSN) (_ *schema.Schema, err error) {
 	if strings.HasPrefix(urlstr, "json://") {
 		return AnalyzeJSON(urlstr)
 	}
+	for _, prefix := range []string{"bq://", "bigquery://", "span://", "spanner://", "dynamodb://", "dynamo://", "mongodb://", "mongo://", "databricks://"} {
+		if strings.HasPrefix(urlstr, prefix) {
+			if err := rejectSSLParams(urlstr); err != nil {
+				return nil, err
+			}
+			break
+		}
+	}
 	if strings.HasPrefix(urlstr, "bq://") || strings.HasPrefix(urlstr, "bigquery://") {
 		return AnalyzeBigquery(urlstr)
 	}
@@ -83,6 +91,14 @@ func Analyze(dsn config.DSN) (_ *schema.Schema, err error) {
 	splitted := strings.Split(u.Short(), "/")
 	if len(splitted) < 2 {
 		return s, fmt.Errorf("invalid DSN: parse %s -> %#v", urlstr, u)
+	}
+
+	changed, err := applySSLParams(u)
+	if err != nil {
+		return nil, err
+	}
+	if changed {
+		urlstr = u.String()
 	}
 
 	opts := []drivers.Option{}
--- a/datasource/datasource.go
+++ b/datasource/datasource.go
@@ -47,6 +47,12 @@ func Analyze(dsn config.DSN) (_ *schema.Schema, err error) {
 		err = errors.WithStack(err)
 	}()
 	urlstr := dsn.URL
+	hasTLS := dsn.TLS != (config.TLS{})
+	if hasTLS {
+		if u, err := dburl.Parse(urlstr); err != nil || !slices.Contains(tlsSupportedDrivers, u.Driver) {
+			return nil, fmt.Errorf("dsn.tls is not supported for this datasource")
+		}
+	}
 	if strings.HasPrefix(urlstr, "https://") || strings.HasPrefix(urlstr, "http://") {
 		return AnalyzeHTTPResource(dsn)
 	}
@@ -55,14 +61,6 @@ func Analyze(dsn config.DSN) (_ *schema.Schema, err error) {
 	}
 	if strings.HasPrefix(urlstr, "json://") {
 		return AnalyzeJSON(urlstr)
-	}
-	for _, prefix := range []string{"bq://", "bigquery://", "span://", "spanner://", "dynamodb://", "dynamo://", "mongodb://", "mongo://", "databricks://"} {
-		if strings.HasPrefix(urlstr, prefix) {
-			if err := rejectSSLParams(urlstr); err != nil {
-				return nil, err
-			}
-			break
-		}
 	}
 	if strings.HasPrefix(urlstr, "bq://") || strings.HasPrefix(urlstr, "bigquery://") {
 		return AnalyzeBigquery(urlstr)
@@ -93,11 +91,10 @@ func Analyze(dsn config.DSN) (_ *schema.Schema, err error) {
 		return s, fmt.Errorf("invalid DSN: parse %s -> %#v", urlstr, u)
 	}
 
-	changed, err := applySSLParams(u)
-	if err != nil {
-		return nil, err
-	}
-	if changed {
+	if hasTLS {
+		if err := applyTLSConfig(u, dsn.TLS); err != nil {
+			return nil, err
+		}
 		urlstr = u.String()
 	}
 
--- a/datasource/datasource.go
+++ b/datasource/datasource.go
@@ -47,8 +47,11 @@ func Analyze(dsn config.DSN) (_ *schema.Schema, err error) {
 		err = errors.WithStack(err)
 	}()
 	urlstr := dsn.URL
-	hasTLS := dsn.TLS != (config.TLS{})
+	hasTLS := dsn.TLS != nil
 	if hasTLS {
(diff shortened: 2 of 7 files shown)


k1LoW:
@mtawil Thank you!!

I think the idea of configuring TLS using common parameters is great.
Since they are common, we might not need to align them with existing MySQL parameters.

TLS configuration sometimes requires files.
Therefore, how about starting by basing it on the configuration in .tbls.yml and translating that into parameters?

e.g;

```yaml
dsn:
  url: my://...
  tls:
    ca: /path/to/ca.pem
    cert: /path/to/client.pem
    key: /path/to/client.key
    verify: identity
```

mtawil:
Thank you @k1LoW for your review!

I agree. The DSN parameters are removed, and TLS is now configured under `dsn.tls` and translated into each driver's native parameters.

```yaml
# .tbls.yml
dsn:
  url: my://dbuser:dbpass@hostname:3306/dbname
  tls:
    ca: /path/to/ca.pem
    cert: /path/to/client.pem
    key: /path/to/client.key
    verify: identity
```

`ca` verifies the certificate chain, `verify: identity` also verifies the hostname, and empty values (e.g. unset env vars) are ignored. Conflicting native TLS parameters in the DSN are rejected with an error.

The diff also fixes an infinite recursion in `DSN.UnmarshalYAML` that crashed any map form config (`dsn: {url: ...}`).

datasource/ssl.go
```diff
@@ -0,0 +1,194 @@
...
+// file must have a .pem or .der extension). go-mssqldb performs full
+// verification (chain + hostname) when encryption is on and
+// trustservercertificate is false, so verify: identity needs no extra mapping.
+func applySQLServerTLS(t config.TLS, values url.Values) error {
+	if t.Cert != "" {
+		return fmt.Errorf("dsn.tls.cert/dsn.tls.key are not supported for driver 'sqlserver'")
+	}
+	if v := values.Get("trustservercertificate"); v != "" {
+		trust, err := strconv.ParseBool(v)
+		if err == nil && trust && t.CA != "" {
```
k1LoW: The guard `t.CA != ""` misses one legitimate conflict.

By the time execution reaches this function, `hasTLS` is guaranteed true, and since `t.Cert != ""` was already rejected above, the only remaining shapes are:

- `ca` only
- `ca` + `verify: identity`
- `verify: identity` only

All three request server certificate verification. `trustservercertificate=true` tells go-mssqldb to skip verification entirely, so it conflicts with all three — including the third case (`verify: identity` alone), which currently slips through and is silently overridden.

One example of a fix:

```go
if err == nil && trust {
    return fmt.Errorf("dsn.tls conflicts with trustservercertificate=true")
}
```

Also worth adding a subcase in `TestApplyTLSConfigSQLServer` for `config.TLS{Verify: "identity"}` + `trustservercertificate=true`, which would have caught this.

datasource/datasource.go
```diff
@@ -47,6 +47,12 @@ func Analyze(dsn config.DSN) (_ *schema.Schema, err error) {
 		err = errors.WithStack(err)
 	}()
 	urlstr := dsn.URL
+	hasTLS := dsn.TLS != (config.TLS{})
```
k1LoW: Using structural equality as "TLS present" is symmetric with `MarshalYAML`'s `d.TLS == (TLS{})` branch, so I read it as intentional. But it does leave one failure mode I would like to close by construction rather than by docs.

Consider:

```yaml
dsn:
  url: my://user:pass@prod-db:3306/app
  tls:
    ca: ${MYSQL_SSL_CA}
```

`config.LoadConfig` runs the input through `expand.ExpandenvYAMLBytes`, which resolves unset env vars to the empty string (`os.LookupEnv` returns `("", false)`, so the placeholder is replaced by nothing). By the time `DSN.UnmarshalYAML` sees the data, `TLS{CA: ""}` equals `TLS{}`, `hasTLS` becomes false, and the connection is made with no TLS enforcement from the `tls:` block — the user wrote `ca:` expecting server verification and would silently get a plaintext (or unverified) handshake instead.

The root cause is that the current shape (`TLS TLS`) cannot distinguish "block absent" from "block present but every field empty". If those two states become distinguishable, an all-empty-but-present block can be rejected at load time.

One example of a fix — make TLS a pointer so "present" travels at the type level, and validate in `Analyze`:

```go
// config/config.go
type DSN struct {
    URL     string            `yaml:"url"`
    Headers map[string]string `yaml:"headers,omitempty"`
    TLS     *TLS              `yaml:"tls,omitempty"`
}
```

```go
// datasource/datasource.go — inside Analyze
if dsn.TLS != nil {
    if *dsn.TLS == (config.TLS{}) {
        return nil, fmt.Errorf("dsn.tls: at least one of ca, cert, key, verify must be set (check that env vars referenced under dsn.tls are set)")
    }
    if u, err := dburl.Parse(urlstr); err != nil || !slices.Contains(tlsSupportedDrivers, u.Driver) {
        return nil, fmt.Errorf("dsn.tls is not supported for this datasource")
    }
}
```

With this, `tls: {}` and `tls: { ca: ${UNSET} }` fail loudly at Analyze time; `tls: { verify: identity }` still works; and omitting `tls:` remains "no TLS", unchanged. `MarshalYAML`'s `d.TLS == (TLS{})` branch becomes `d.TLS == nil` and stays clean.

Other shapes would work too — for example a `TLS.UnmarshalYAML` that rejects an empty block, or an unexported presence bit — but the pointer route keeps parsing and validation cleanly separated. Happy to go with whatever you prefer; the key point is closing the fail-open path, not the exact mechanism.

config/config.go
```diff
@@ -69,6 +69,15 @@ type Config struct {
...
 	Headers map[string]string `yaml:"headers,omitempty"`
+	TLS     TLS               `yaml:"tls,omitempty"`
+}
+
+// TLS is the SSL/TLS configuration for the datasource connection.
+type TLS struct {
+	CA     string `yaml:"ca,omitempty"`
+	Cert   string `yaml:"cert,omitempty"`
+	Key    string `yaml:"key,omitempty"`
+	Verify string `yaml:"verify,omitempty"`
```
k1LoW, line 80: Nit / scope note, not blocking: `verify` currently accepts only `"identity"` (and empty). It is a natural extension axis — `off` / `preferred` / `encrypt` / `ca` / `identity` could all live on the same field, similar to postgres `sslmode`, and would let `dsn.tls` express the full range of TLS modes without having to fall back to native DSN params for the non-verifying cases.

I think extending the enum is best done in a follow-up PR rather than in this one — this PR's scope (introducing `dsn.tls` with the verifying subset) is already the right size. Just leaving a marker here so the design intent ("verify is an enum, expected to grow") is visible when the next iteration happens.

mtawil: Agreed. Treating `verify` as a growing enum (`off` / `preferred` / `encrypt` / `ca` / `identity`) is the natural direction, and keeping it out of this PR keeps the scope right. Happy to pick that up in a follow-up PR once this one lands.

k1LoW:
I suspect that the lint error is caused by https://github.com/golangci/golangci-lint/issues/6732.

k1LoW:
@mtawil Great!! Thank you!!

PR #851, opened 2026-07-30
fix: stabilize trigger introspection order in postgres/mysql/mssql/sqlite

### LINKED ISSUE #850: Make trigger output deterministic in postgres/mysql/mssql/sqlite
## Background

The introspection queries do not fix the order of triggers, so **the same schema can produce different documents between runs** — the order is whatever access path the database happens to choose. With the standard tbls workflow (commit the generated documents, review them as diffs) this shows up as row reordering unrelated to any schema change, and when generation runs in CI, local runs and CI runs keep rewriting the order back and forth. In one project a table with 3 triggers had its Triggers section flip between two orders 4 times, with no schema change at all.

Same class of problem as #829 (MySQL constraints), fixed in #830 by giving the query an explicit `ORDER BY`.

**postgres** has an `ORDER BY`, but it is not a sort key (`drivers/postgres/postgres.go`, in `Analyze`):

```sql
SELECT tgname, pg_get_triggerdef(trig.oid), descr.description AS comment
FROM pg_trigger AS trig
LEFT JOIN pg_description AS descr ON trig.oid = descr.objoid
WHERE tgisinternal = false
AND tgrelid = $1::oid
ORDER BY tgrelid
```

`tgrelid` is pinned to a single value by the `WHERE` clause, so every row carries the same sort key and the order is left unspecified: a Seq Scan returns heap physical order, an Index Scan on `pg_trigger_tgrelid_tgname_index` returns `tgname` order. Both plans occur in normal use — in one measurement the same session used a Seq Scan for one table and an Index Scan for another. The constraints and indexes queries in the same file are totally ordered; triggers are the exception.

The trigger queries in `drivers/mysql/mysql.go` (`information_schema.triggers`), `drivers/mssql/mssql.go` (`sys.triggers`) and `drivers/sqlite/sqlite.go` (`sqlite_master`) have no `ORDER BY` at all. Checked against v1.95.0.

<details>
<summary>Reproduction (PostgreSQL 18.3)</summary>

Three triggers on one table, created as `z_third` → `m_second` → `a_first`. Running the driver's own query twice, switching only the plan:

```sql
SET enable_indexscan = off; SET enable_bitmapscan = off;   -- Seq Scan
SELECT tgname FROM pg_trigger WHERE tgisinternal = false
AND tgrelid = 'demo'::regclass::oid ORDER BY tgrelid;
--> z_third, m_second, a_first

SET enable_seqscan = off; SET enable_indexscan = on;       -- Index Scan, same query
--> a_first, m_second, z_third
```

The `enable_*` settings only make the reproduction reliable; both plans occur without them.

Heap order is not stable either: after dropping and re-creating one of three triggers under the same name, its row moved to the end of the heap (`ctid` (7,49) → (7,51)), so a Seq Scan would return `t_a, t_c, t_b` instead of `t_a, t_b, t_c`. Even with an identical current schema, the history of how it was built leaks into the output.

</details>

## Requirements

- Deterministic per-table trigger order in the 4 drivers that read triggers: **postgres / mysql / mssql / sqlite**
- Ordered by **trigger name, ascending**
- Independent of the query plan, statistics, and how the schema was built

No other driver needs changing: `mariadb` embeds `mysql.Mysql` and follows automatically, `redshift` skips trigger collection (`if !p.rsMode`), and the rest never touch triggers.

Name order is chosen because PostgreSQL fires triggers of the same timing and event alphabetically, so there it also matches the firing order. The other three do not guarantee that, but aligning all four keeps them consistent; representing firing order is not a goal here.

## Acceptance criteria

- [ ] For a table with multiple triggers, the generated Triggers section is ordered by trigger name ascending, in all four drivers
- [ ] For postgres, the output is identical regardless of how `enable_seqscan` / `enable_indexscan` / `enable_bitmapscan` are set
- [ ] A database where a trigger was dropped and re-created under the same name produces the same output as one built from scratch

A total order in the query guarantees the last two structurally, so checking them manually is enough — no need to drive planner settings from the test suite.

Note that this also reorders the committed `sample/sqlite/posts.md` and `sample/sqlite/schema.json`, since `testdata/ddl/sqlite.sql` already puts 4 triggers on `posts` — the samples need regenerating in the same change.

## Out of scope

- Ordering of schema elements other than triggers — postgres constraints / indexes are already totally ordered; other introspected elements such as functions are uninvestigated and belong in a separate issue
- Making trigger order configurable — `format.sort` already exists for users who want everything sorted

---

I can send a PR for this if the direction looks reasonable.

178inaba:
## Why

The trigger queries do not fix the order of their results, so the same schema can produce different documents between runs — the order is whatever access path the database happens to choose. For the standard tbls workflow (commit the generated documents, review them as diffs) this shows up as row reordering unrelated to any schema change, and when generation runs in CI, local runs and CI runs keep rewriting the order back and forth.

- **postgres** has an `ORDER BY`, but `tgrelid` is pinned to a single value by the `WHERE` clause, so every row carries the same sort key and the ordering is left unspecified. A Seq Scan returns heap physical order, an Index Scan on `pg_trigger_tgrelid_tgname_index` returns `tgname` order — and both plans occur in normal use
- **mysql / mssql / sqlite** have no `ORDER BY` at all

This is the same class of problem as #829 (MySQL constraints), fixed in #830 by giving the query an explicit `ORDER BY`.

Closes #850

## What

Give the trigger query in each of the four drivers a total order, following the ordering convention each driver already uses:

| Driver | Change | Key |
| --- | --- | --- |
| postgres | `ORDER BY tgrelid` → `ORDER BY trig.oid` | creation order |
| mssql | add `ORDER BY t.object_id` | creation order |
| sqlite | add `ORDER BY rowid` | creation order |
| mysql | add `ORDER BY event_object_table, trigger_name` | name |

`TestTriggersOrder` is added to all four driver test packages. postgres / mysql / mssql create a table whose triggers are created in descending name order, so creation order and name order differ, then assert the expected order and drop the table — the same shape as `TestCheckConstraintsOrder` from #830. sqlite asserts the four triggers already on `posts` in `testdata/ddl/sqlite.sql`, without writing to the shared test database that `make testdoc_sqlite` also reads.

No sample regeneration is needed: `sample/sqlite/posts.md` is the only file under `sample/` with more than one trigger on a table, and it already records them in creation order. The diff is limited to `drivers/`.

No other driver needs changing: `mariadb` embeds `mysql.Mysql` and follows automatically, `redshift` skips trigger collection via `if !p.rsMode`, and the rest never touch triggers.

## Notes for reviewers

**Ordering follows each driver's existing convention rather than being uniformly alphabetical**, per the direction in [this comment](https://github.com/k1LoW/tbls/issues/850#issuecomment-5139933064). tbls orders by definition/creation order in postgres (`oid`, `attnum`, `indexrelid`) and mssql (`OBJECT_ID`, `column_id`, `index_id`), and sqlite follows definition order through `PRAGMA`. Making triggers alone alphabetical would have introduced an approach that contradicts the surrounding code in those drivers, so consistency within each driver was prioritised over byte-identical order across DBMSs.

**mysql is the exception and stays name-ordered**, because `information_schema` has no stable key equivalent to an oid. This is the same constraint that made #830 alphabetical. Its first sort key, `event_object_table`, does not affect the output — the query fetches the triggers for the whole schema at once and the rows are bucketed into a per-table map afterwards — but it is kept so the query keeps the same shape as the constraint query fixed in #830.

**Two of the acceptance criteria in #850 no longer hold as written**, and the issue body has been left as the original record rather than rewritten:

- Criterion 1 said "ordered by trigger name ascending". That reads as "totally ordered" now — the order is deterministic and plan-independent, but it is creation order in three of the four drivers
- Criterion 3 (a database where a trigger was dropped and re-created under the same name produces the same output as one built from scratch) is not met. As noted in the comment above, this is already the case for tables and indexes, so it is not a deficiency specific to triggers

Criterion 2 (output independent of `enable_seqscan` / `enable_indexscan` / `enable_bitmapscan`) is met: the sort key is unique per row, so the total order is structural and holds under any plan.

**What the tests do and do not cover**: they pin the expected order, so a regression that reintroduces name sorting — or any other reordering — fails them. They are not a guard against the `ORDER BY` being deleted outright, since the natural scan order often coincides with creation order. That follows from #850's own note that a total order in the query guarantees the plan-independence structurally, so checking it manually is enough and there is no need to drive planner settings from the test suite.

One consequence worth naming: in PostgreSQL, triggers of the same timing and event fire in alphabetical order, so name order would have doubled as firing order there. Creation order does not. Representing firing order was never a goal of #850, and the other three DBMSs make no such guarantee, so this seemed an acceptable loss for driver-internal consistency.

<details>
<summary>sqlite: measured behaviour of <code>rowid</code> in <code>sqlite_master</code></summary>

Re-applying `testdata/ddl/sqlite.sql` to an existing database (what `make db_sqlite` does) keeps the relative order — the rowids are renumbered but the sequence is unchanged:

```
first build:   18|update_posts_updated  59|posts_fts_insert  60|posts_fts_delete  61|posts_fts_update
re-applied:    75|update_posts_updated  93|posts_fts_insert  94|posts_fts_delete  95|posts_fts_update
```

`VACUUM` compacts the rowids but preserves the order. Dropping and re-creating a trigger under the same name moves it to the end, which is the criterion 3 case above.

</details>

## Verification

All four drivers were verified locally. For postgres / mssql / sqlite the test was confirmed to fail before the query change and pass after it — before the change each returned the triggers in name order:

```
$ go test ./drivers/postgres/... ./drivers/mysql/... ./drivers/mssql/... ./drivers/sqlite/... \
    -tags 'postgres mysql mssql sqlite sqlite_fts5'
ok  github.com/k1LoW/tbls/drivers/postgres
ok  github.com/k1LoW/tbls/drivers/mysql
ok  github.com/k1LoW/tbls/drivers/mssql
ok  github.com/k1LoW/tbls/drivers/sqlite

$ make testdoc_sqlite     # passes without regenerating the sample
$ ./tbls diff pg://... sample/postgres     # no diff
$ ./tbls diff my://... sample/mysql        # no diff
$ ./tbls diff ms://... sample/mssql        # no diff

$ golangci-lint run --build-tags 'postgres mysql mssql sqlite sqlite_fts5' ./drivers/...
0 issues.
```

### DIFF
--- a/drivers/mssql/mssql.go
+++ b/drivers/mssql/mssql.go
@@ -331,6 +331,7 @@ INNER JOIN sys.sql_modules AS sm
 ON sm.object_id = t.object_id
 WHERE type = 'TR'
 AND parent_id = object_id(@p1)
+ORDER BY t.object_id
 `, fmt.Sprintf("%s.%s", tableSchema, tableName))
 		if err != nil {
 			return errors.WithStack(err)
--- a/drivers/mssql/mssql_test.go
+++ b/drivers/mssql/mssql_test.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/k1LoW/tbls/schema"
 	_ "github.com/microsoft/go-mssqldb"
 	"github.com/xo/dburl"
@@ -58,6 +59,45 @@ func TestInfo(t *testing.T) {
 	}
 	if d.DatabaseVersion == "" {
 		t.Errorf("got not empty string.")
+	}
+}
+
+func TestTriggersOrder(t *testing.T) {
+	if _, err := db.Exec(`DROP TABLE IF EXISTS tbls_trigger_order`); err != nil {
+		t.Fatal(err)
+	}
+	if _, err := db.Exec(`CREATE TABLE tbls_trigger_order (a int)`); err != nil {
+		t.Fatal(err)
+	}
+	defer func() {
+		_, _ = db.Exec(`DROP TABLE IF EXISTS tbls_trigger_order`)
+	}()
+	// Create the triggers in descending name order so that creation order and name order differ.
+	// CREATE TRIGGER must be the first statement in a batch, so each one is executed on its own.
+	if _, err := db.Exec(`CREATE TRIGGER trg_tbls_trigger_order_b ON tbls_trigger_order AFTER INSERT AS BEGIN SET NOCOUNT ON; END`); err != nil {
+		t.Fatal(err)
+	}
+	if _, err := db.Exec(`CREATE TRIGGER trg_tbls_trigger_order_a ON tbls_trigger_order AFTER INSERT AS BEGIN SET NOCOUNT ON; END`); err != nil {
+		t.Fatal(err)
+	}
+
+	driver := New(db)
+	sc := &schema.Schema{Name: "testdb"}
+	if err := driver.Analyze(sc); err != nil {
+		t.Fatal(err)
+	}
+
+	tbl, err := sc.FindTableByName("tbls_trigger_order")
+	if err != nil {
+		t.Fatal(err)
+	}
+	var got []string
+	for _, trig := range tbl.Triggers {
+		got = append(got, trig.Name)
+	}
+	want := []string{"trg_tbls_trigger_order_b", "trg_tbls_trigger_order_a"}
+	if diff := cmp.Diff(got, want); diff != "" {
+		t.Error(diff)
 	}
 }
 
--- a/drivers/mysql/mysql.go
+++ b/drivers/mysql/mysql.go
@@ -177,6 +177,7 @@ SELECT
   action_statement
 FROM information_schema.triggers
 WHERE event_object_schema = ?
+ORDER BY event_object_table, trigger_name
 `, s.Name)
 	if err != nil {
 		return errors.WithStack(err)
--- a/drivers/mysql/mysql_test.go
+++ b/drivers/mysql/mysql_test.go
@@ -8,6 +8,7 @@ import (
 	"testing"
 
 	_ "github.com/go-sql-driver/mysql"
+	"github.com/google/go-cmp/cmp"
 	"github.com/k1LoW/tbls/schema"
 	"github.com/xo/dburl"
 )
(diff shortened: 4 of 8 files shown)


178inaba:
Force-pushed to follow the direction given in [k1LoW/tbls#850 (comment)](https://github.com/k1LoW/tbls/issues/850#issuecomment-5139933064): the sort key now matches the ordering convention each driver already uses, instead of being uniformly alphabetical.

- postgres `ORDER BY trig.oid` / mssql `ORDER BY t.object_id` / sqlite `ORDER BY rowid` (creation order)
- mysql stays on `ORDER BY event_object_table, trigger_name`, since `information_schema` has no oid equivalent
- `sample/sqlite/` no longer needs regenerating — it already records the triggers in creation order — so the diff is limited to `drivers/`

The PR description has been rewritten accordingly, and the reasoning about the acceptance criteria is in the issue thread.

k1LoW:
@178inaba Great!! Thank you!!

PR #856, opened 2026-08-15
fix: --adjust-table misaligns columns containing East Asian characters

VXNCXNX:
## What's broken

`--adjust-table` misaligns any column containing East Asian characters.

`adjustTable` measures column width with `runewidth.StringWidth`, which counts display cells, then pads with `%-Ns`, which counts runes. For a character that occupies two cells the two disagree, so the cell is padded to N runes and ends up 2N cells wide.

Padding to a target of 8, with the two approaches side by side:

```
"ユーザー"    %-8s -> width 12    FillRight -> width 8
"id"         %-8s -> width  8    FillRight -> width 8
"説明"        %-8s -> width 10    FillRight -> width 8
```

The whole point of `--adjust-table` is that the raw markdown lines up in a text editor, so a Japanese or Chinese column name defeats it for every row of that table.

## The fix

`runewidth.FillRight`, from the package already imported two lines above to do the measuring. That makes the padding use the same notion of width as the measurement.

ASCII is unaffected, since `FillRight` and `%-Ns` agree when every rune is one cell wide. The existing `--adjust` golden tests pass unchanged, which is the check that this does not disturb the ASCII fixtures.

## Verification

`TestAdjustTableEastAsianWidth` in `output/md/md_test.go` runs a table mixing `ユーザー`, `説明` and ASCII through `adjustTable`, then asserts every cell in a column has the same display width as its header. That is the property `--adjust-table` exists to provide, rather than a golden of one particular string.

Reverting only the padding expression fails it:

```
md_test.go:271: column 0 row 2: got display width 12, want 8 ("ユーザー    ")
md_test.go:271: column 1 row 3: got display width 9, want 7 ("説明     ")
```

`go test ./output/...` passes, `gofmt -l output/` is clean and `go vet ./output/...` is clean.

*Disclosure: written with AI assistance (Claude Code). I confirmed the width difference by running both padding approaches directly, and ran the revert check myself.*

### DIFF
--- a/output/md/md.go
+++ b/output/md/md.go
@@ -964,7 +964,7 @@ func adjustTable(data [][]string) [][]string {
 			if i == 1 {
 				data[i][j] = strings.Repeat("-", w[j])
 			} else {
-				data[i][j] = fmt.Sprintf(fmt.Sprintf("%%-%ds", w[j]), r.Replace(data[i][j]))
+				data[i][j] = runewidth.FillRight(r.Replace(data[i][j]), w[j])
 			}
 		}
 	}
--- a/output/md/md_test.go
+++ b/output/md/md_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/k1LoW/tbls/config"
 	"github.com/k1LoW/tbls/schema"
 	"github.com/k1LoW/tbls/testutil"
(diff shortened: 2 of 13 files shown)


k1LoW (approved):
Thank you!!

PR #857, opened 2026-08-25
feat(xlsx): link table names in the index sheet to their sheets

### LINKED ISSUE #506: Add Hyperlinks to "Tables of" Sheet for Sheets in Excel Output
Thanks for the very useful tbls!

I'd like to suggest a feature that I think would make it even better. How about adding hyperlinks in the Excel output's "Tables of" sheet, leading directly to individual table sheets? For example, if there's a table named "users", a hyperlink in the "Tables of" sheet could look something like this: 
`=HYPERLINK("#users!A1","users")`

I believe this would make navigation smoother and the documentation more user-friendly in Excel.

Here's a screenshot to illustrate what I mean:

image

Thanks again for your hard work on tbls and for considering this feature!

ChrisJr404:
Closes #506.

The "Tables of" index sheet in the xlsx output lists every table name, but they were plain text. This turns each name into a `HYPERLINK("#'<sheet>'!A1", "<name>")` formula so you can click a table in the index and jump straight to its sheet. Handy once a schema has more than a handful of tables.

A couple of small things to keep the links valid:

The target sheet name is single quoted, so names with spaces still resolve, and I pulled the existing 31 character truncation out into a `tableSheetName` helper that both the index link and the table sheet use, so a long name links to the same sheet it actually creates. Any embedded quote gets doubled per Excel's rules.

I added a test that renders a schema (including a name with a space and one over 31 chars), unzips the result and checks the formulas land in the index sheet. There was no test file for this package before.

### DIFF
--- a/output/xlsx/xlsx.go
+++ b/output/xlsx/xlsx.go
@@ -121,7 +121,7 @@ func (x *Xlsx) createSchemaSheet(w *excl.Workbook, s *schema.Schema) error {
 	})
 	n := 5
 	for i, t := range s.Tables {
-		setStringWithBorder(sheet, n+i, 1, t.Name)
+		setFormulaWithBorder(sheet, n+i, 1, hyperlinkToSheet(tableSheetName(t.Name), t.Name))
 		setNumberWithBorder(sheet, n+i, 2, len(t.Columns))
 		setStringWithBorder(sheet, n+i, 3, t.Comment)
 		setStringWithBorder(sheet, n+i, 4, t.Type)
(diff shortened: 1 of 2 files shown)


k1LoW:
@ChrisJr404 Thank you for the PR!

It was an issue that had been left untouched for three years, so this is a huge help. I appreciate that you even wrote tests for it.

I have just one question. When I looked at your account, I saw that you've submitted PRs to hundreds of repositories in just a few months, which seems beyond human capability (I'm guessing you might be using AI to help). I have no intention of blaming you at all. I'm just genuinely curious. Why, and through what mechanism, are you keeping this up?
I'd especially love to know the "why."

Feel free to answer only as much as you're comfortable with 👍

